Issue #11, opened 2026-07-10
[C55] Central reusable claude-run.yml can't replace the four repos' local copies — it lacks their runner, prompt, tooling, and bug-fix behaviors


richkuo:
**Complexity: 55/100** — scope: large (one workflow redesigned around ~6 new inputs/behaviors, then four downstream migrations); risk: medium-high (every repo's @claude automation rides on it; prompt-safety regressions are silent); uncertainty: medium (prompt-override design and reusable-workflow input limits need working out)

## Problem

The centralized reusable workflow (`.github/workflows/claude-run.yml`, published for cross-repo use and referenced by `templates/claude-workflow/workflows/claude.yml`) was meant to replace the local copies in go-trader, scene-cut-ai, scenecut-front, and job-search. All four migration attempts were correctly aborted (richkuo/go-trader#1311, richkuo/scene-cut-ai#130, richkuo/scenecut-front#92, richkuo/job-search#288 — each PR body carries its repo's full reconciliation list) because the local copies have diverged with behavior the central workflow cannot express:

1. **Runner**: all four run `runs-on: self-hosted`; central hardcodes `ubuntu-latest` with no override input.
2. **Prompt overrides are append-only**: central stages its own `templates/claude-workflow/prompts/*.md` and only appends `.github/prompts/<name>-local.md`. Repos have *replaced* base prompt text with safety-hardened variants (go-trader: trading-safety carve-outs, backtest bans, gofmt; scenecut-front: Prisma/DB-destructive-change bans; scene-cut-ai: Expo/no-test-suite guidance). Append can't remove or rewrite base text, so switching silently drops safety language.
3. **Per-repo tooling**: go-trader and job-search set up Go and allowlist `gofmt`; central has no toolchain setup or allowlist-extension input.
4. **Missing flow**: job-search supports a `sync-release` flow; central only has `sync-docs`/`create-release`.
5. **Local bug-fixes not upstreamed**: PR-head-SHA change detection via `gh pr view --json headRefOid` (central's `git rev-parse HEAD` false-positives against the action's own checkout); a "Fail if Claude Code returned an error" execution-file check; go-trader's implement-mode routing and 90-minute timeout.

## Goal

A repo can adopt the central workflow with zero behavior loss: every divergence above is expressible via inputs or per-repo files, and the local bug-fixes live in the central copy.

## Approach

[...]

PR #12, opened 2026-07-10
Central claude-run.yml: inputs, replace-capable prompts, sync-release, upstreamed fixes (#11)

richkuo:
## Summary

Implements the central-workflow half of #11: every divergence that blocked the four consumer repos (go-trader, scene-cut-ai, scenecut-front, job-search) from adopting the reusable workflow is now expressible via inputs or per-repo files.

### New `workflow_call` inputs
| Input | Default | Covers |
|-------|---------|--------|
| `runs_on` | `ubuntu-latest` | job-search / scene-cut-ai / scenecut-front self-hosted runners |
| `timeout_minutes` | `45` | go-trader's 90-minute implement runs |
| `go_version` | (empty) | go-trader / job-search Go setup — gofmt-only, opt-in |
| `extra_allowed_tools` | (empty) | `Bash(gofmt *)` etc.; appended to every route EXCEPT the hard-locked create-release allowlist, validated with the same shell-safety character rule as prompts |

### Prompt sourcing: replace-or-append
`.github/prompts/<name>.md` in the consumer repo now replaces the shared base prompt (needed for the repos' safety-hardened variants: trading-safety carve-outs, Prisma/DB bans, Expo guidance); `<name>-local.md` still appends to whichever base was chosen. Both go through the existing shell-safety check.

### sync-release flow
Accepted by the `flow` input; the shared sync-docs allowlist gains `docs-release/*` branch/push globs, and the shared `sync-docs-release.md` prompt gains a generic sync-release procedure (job-search's repo-specific bits stay in its replacement prompt). The vendored `claude.yml` template does not detect the phrase by default — documented opt-in.

### Upstreamed bug fixes
- PR-head-SHA change detection via `gh pr view --json headRefOid` (the local `git rev-parse HEAD` before/after comparison false-positives against claude-code-action's own checkout).
- `Fail if Claude Code returned an error` execution-file check, so an API error mid-run fails the job instead of passing silently.

### Docs
Template README gains a Customization inputs section and the replace/append override description; the template `claude.yml` comment explains sync-release opt-in.

## Verification
- `js-yaml` parses both edited workflow files.
- Full template test suite: 73 tests, OK (includes `test_workflow_logic.py`, which executes the real classifier shell from `claude.yml`).
- Prompts re-checked for the forbidden `\"` / backtick / \$ characters: clean.

Part of #11 — the four per-repo migration PRs (delete local `claude-run.yml`, pass inputs, keep replaced prompts) follow separately, one repo at a time.

---

https://claude.ai/code/session_01RCjSkFmes3abQRRZarweF5

### DIFF
--- a/.github/workflows/claude-run.yml
+++ b/.github/workflows/claude-run.yml
@@ -7,10 +7,16 @@ name: Claude Code run
 # with secrets: inherit. Prompt files and comment-patch scripts are fetched
 # from rk-skills at run time (templates/claude-workflow/), so updating
 # rk-skills updates every consumer repo on its next run with no per-repo sync.
-# A consumer repo appends route-specific tailoring by adding
-# .github/prompts/<prompt-name>-local.md (e.g. fix-pr-review-local.md), whose
-# text is appended to the shared prompt and validated by the same
-# shell-safety check.
+# A consumer repo tailors prompts two ways (both validated by the same
+# shell-safety check):
+#   - replace: .github/prompts/<prompt-name>.md (e.g. issue-workflow.md)
+#     substitutes for the shared prompt entirely — for repos whose base prompt
+#     carries safety-hardened language the shared text lacks.
+#   - append:  .github/prompts/<prompt-name>-local.md (e.g.
+#     fix-pr-review-local.md) is appended to whichever base was chosen.
+# Runner, timeout, toolchain, and allowlist divergences are expressed via the
+# optional runs_on / timeout_minutes / go_version / extra_allowed_tools inputs
+# (issue #11), so consumer repos need no local fork of this file.
 # The GITHUB_TOKEN permissions are NOT declared here on purpose: a called
 # workflow inherits the caller job's permissions, so the least-privilege review
 # job and the write-capable implement job differ only at the call site.
@@ -41,7 +47,7 @@ on:
         required: true
         type: string
       flow:
-        description: 'Docs/release flow: empty, sync-docs, or create-release'
+        description: 'Docs/release flow: empty, sync-docs, create-release, or sync-release'
         required: false
         type: string
         default: ''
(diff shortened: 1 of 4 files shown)


richkuo:
@claude review

richkuo:
Needs Updates

### Needs Fixing

1. **Repo prompt overrides can be attacker-controlled on PR-review events.**
`.github/workflows/claude-run.yml:290-293` reads `.github/prompts/<name>.md` (and the pre-existing `-local.md` append) from the workspace checkout. On `pull_request_review` / `pull_request_review_comment` events, `actions/checkout` defaults to the PR merge commit — so a fork PR shipping `.github/prompts/pr-review-format.md` fully replaces the review system prompt when a maintainer triggers `@claude review`, letting an untrusted diff dictate its own review (e.g. force `LGTM`). The append vector predates this PR, but replacement completes the takeover. Source overrides from the trusted default branch (`git show origin/<default_branch>:.github/prompts/...`) instead of the working tree.
**Invariant:** untrusted PR content must never influence the agent's system prompt or capabilities.
**Must survive:** fork PR adds/edits either override file and a maintainer requests review; PR deletes the override that main has; issue-triggered runs where the checkout is main.

2. **The CLAUDE.md-revise pass becomes unreachable.**
The new detection (`claude-run.yml:369-380`) returns `made_commits=false` whenever there's no PR baseline — but under the template's routing, implement mode only ever fires from non-PR triggers (plain issues, docs flows), so the baseline is always empty and the revise step never runs, silently retiring the feature for every migrating repo except job-search (where it's already dead). Use go-trader's hybrid instead: PR-head SHA via API when a PR resolves, else local `git rev-parse HEAD` comparison — that still fixes the comment-triggered false positive while keeping docs-sync's in-place commits detectable.
**Invariant:** upstreaming a bug fix must not disable an existing feature for repos that don't have that bug.
**Must survive:** docs-sync run that commits (revise runs); issue-workflow run committing only inside a linked worktree (revise skipped); comment-triggered PR run (no false positive).

### Recommended Optional

3. **The error-check step assumes `jq` exists on the runner.**
`claude-run.yml:337-352` hard-fails every run if `jq` is absent, and three target repos use self-hosted runners whose tooling I can't verify. Guard with `command -v jq` (warn-and-skip) or document `jq` as a runner prerequisite in the template README.

---
Validated with LLM: Fable 5 | medium | Harness: Claude Code

richkuo:
Review dispositions (commit e8747f9):

1. **Attacker-controlled prompt overrides — Fixed.** Both override forms (replace and `-local.md` append) are now read from the repository's default branch (`git fetch refs/heads/<default_branch>` + `git show FETCH_HEAD:...`), never the working tree. Verified with a simulated fork-PR attack: attacker files in the checkout are ignored, the default-branch text wins, and an attacker-only `-local.md` is not appended. If the default branch cannot be fetched, overrides are skipped with a warning (fail-safe to the shared prompt). This also closes the pre-existing append-only vector.

2. **Revise pass unreachable — Fixed.** Baseline is now hybrid: PR-head SHA via `gh pr view --json headRefOid` when the trigger resolves to a PR (no checkout false positives), otherwise local HEAD comparison plus the original dirty-tree check — so docs-sync and other in-place commits keep triggering the CLAUDE.md revise pass, and worktree-only issue runs still skip it.

3. **jq assumption — Fixed.** The error check now skips with a `::warning` when `jq` is absent; the template README documents that self-hosted runners should provide `git`, `gh`, and `jq`.

Re-verified: YAML parses, 73 template tests pass, prompts still clean of shell-unsafe characters.

@claude review

---
Updated with LLM: Fable 5 | medium | Harness: Claude Code

richkuo:
LGTM

---
Validated with LLM: Fable 5 | medium | Harness: Claude Code

PR #36, opened 2026-07-13
fix: authorize the milestone plan comment so it stops tripping the safety classifier

richkuo:
## Problem

A `milestone-pipeline` run completed correctly, but the harness safety classifier raised an `[External System Writes]` warning against the Plan stage: the planning agent posted an implementation-plan comment on an issue the session had not created, with "no user message authorizing the posting".

The flagged behavior is not a bug — posting the plan to the issue **is** the Plan stage's designed output. The comment is the handoff artifact the builder implements against, and both the workflow's `phases` metadata ("plans posted to the issues") and the skill's step 4 describe it.

## Why it fired

The classifier evaluates each subagent's actions against the prompt that agent was given. `planPrompt()` said only *"Post the plan as a comment on issue #N"* — an instruction to write to an external system, with nothing in the agent's context tying that write to a user authorization. The classifier could not distinguish an authorized, designed-for artifact from an unprompted publish under the user's identity, so it warned. Its own stated clearing condition (the user authorizing a comment on that issue) was in fact satisfied upstream — the user approves the run plan at milestone-workflow step 2 — but that approval was never surfaced into the agent's prompt.

A false positive with a real root cause: **the authorization existed but was invisible to the agent doing the writing.**

## Fix

Make the authorization both genuine and visible, at the two points where it matters.

- **`skills/milestone-workflow/SKILL.md` step 2** — the mandatory run-plan review now requires stating the GitHub writes the run performs (issue-body corrections, plan comments, PRs, review-trigger and disposition comments), and that merging still stays with the user. The user's approval now covers those writes explicitly instead of by implication.
- **`workflows/milestone-pipeline.js` `planPrompt()`** — the planning agent is now told that the user approved this run plan, that the approval authorizes commenting the plan on the issue, and that the comment is the purpose of the step rather than an incidental side effect. The negative bound is tightened in the same breath: do not modify files, **do not comment anywhere else**, do not start implementing.

The narrowed negative bound matters as much as the authorization: the agent is cleared to write exactly one comment, on exactly the issue it was dispatched for.

## Verification

`bun test` — 37 pass, 0 fail. No change to the workflow's control flow; the diff is prompt and documentation text only.

---
Created with LLM: Opus 4.8 | high | Harness: Claude Code


richkuo:
@claude review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-07-13T23:19:32Z) in 7165d21.

### Fixed (recommended optional)

1. **Expose authorization for validation-driven issue edits** — the conditional correction block now states that the approved milestone run authorizes applying corrections, while the zero-corrections path emits no authorization language; regression coverage verifies both paths (`workflows/milestone-pipeline.js:260`, `tests/milestone-pipeline.test.js:358`).

---
Created with LLM: GPT-5 | high | Harness: Codex

richkuo:
@claude sonnet review

PR #42, opened 2026-07-15
Teach fix-pr-review to fix failing CI checks

richkuo:
## Summary
- `fix-pr-review` currently only resolves review comments (formal reviews, issue comments, inline diff threads) — it ignores failing CI checks entirely.
- Adds a new step 1.5 that takes a one-shot `gh pr checks` snapshot, skips any check that hasn't finished (`pending`/`skipping`/unresolved `cancel`) rather than waiting or polling, and turns each `fail`-bucket check into a `CI Failure` finding that flows through the existing validate → fix → verify → disposition pipeline.
- Validation for CI findings is spelled out separately: a failure is only fixed if it traces to this PR's diff; a failure that's pre-existing on the base branch or a one-off flake is refuted with evidence and surfaced to the user instead of patched around.
- Updated the disposition comment template, red-flags table, and common-mistakes list to cover the new source.

## Test plan
- [ ] Read through the updated `skills/fix-pr-review/SKILL.md` for internal consistency (step numbering, terminology, template sections)
- [ ] Run `/fix-pr-review` on a PR with a genuinely failing CI check and confirm it fixes the root cause without waiting on other in-progress checks
- [ ] Run `/fix-pr-review` on a PR with a pre-existing/flaky failing check and confirm it's refuted rather than patched around

---
Created with LLM: Sonnet 5 | high | Harness: Claude Code


richkuo:
@claude review

richkuo:
@claude review

richkuo:
Addressed review feedback (Opus 4.8 review bot · 2026-07-15T09:12:35Z) in 2db9f1a.

### Fixed
1. **CI terminology mismatch between step 1.5's bucket logic and the summary rows** — Red Flags row and Common Mistakes bullet used raw check-run status words (`in_progress`/`queued`/`completed`) that `gh pr checks --json` never emits, and "skip anything not `completed`" contradicted 1.5's own logic for `skipping`/`cancel` buckets. Switched both to the `bucket` vocabulary (`pending`/`skipping`) step 1.5 actually uses, and reworded the channel-count language to reference steps 1 and 1.5 directly instead of a "three vs four" count (`skills/fix-pr-review/SKILL.md:247-248,264-265`).

### Not changed (refuted)
1. **`gh pr checks` non-zero exit on failure** — verified empirically on a disposable probe PR (#44, closed): the bare `gh pr checks <N>` does exit 1 on a failing check, but the exact command step 1.5 uses — `gh pr checks <N> --json ...` — exits **0** even when a check's bucket is `fail`. The `--json` flag changes the exit-code behavior, so the scenario the finding warns about doesn't occur for this skill's actual invocation. No wording change needed.

---
Created with LLM: Sonnet 5 | high | Harness: Claude Code

richkuo:
@claude sonnet review

richkuo:
Addressed review feedback (Sonnet 5 review bot · 2026-07-15T09:22:37Z) in a2e31b7.

### Fixed
1. **`{sha}` placeholder in the external-CI fallback command is never auto-filled by `gh api`** — confirmed via `gh api --help`: only `{owner}`/`{repo}`/`{branch}` are auto-substituted. Added an explicit `gh pr view <N> --json headRefOid --jq .headRefOid` step and switched the placeholder to `<sha>`, matching the doc's existing manually-filled-value convention (`skills/fix-pr-review/SKILL.md:79`).
2. **No detail-fetching procedure for a `cancel`-bucket check caused by a genuine upstream failure** — the cancelled run's own log reports `cancelled` steps, not `failure`, so the existing `fail`-bucket command wouldn't surface anything there. Added guidance to pull detail from the upstream check that actually failed via the `fail`-bucket procedure, and cite that check's name as the finding (`skills/fix-pr-review/SKILL.md:76`).

---
Created with LLM: Sonnet 5 | high | Harness: Claude Code

richkuo:
@claude review

richkuo:
Addressed review feedback (Opus 4.8 review bot · 2026-07-15T09:41:22Z) in a990738.

### Fixed
1. **`cancel`-bucket redirect duplicates a `fail`-bucket finding** — confirmed: the redirect unconditionally cited the upstream failing job's name, but that job is itself a `bucket: fail` entry the fail-bucket path already turns into its own CI Failure finding (matrix fail-fast / `needs:`-gated case). The cancelled check now first checks the same snapshot for that `fail`-bucket entry and skips itself entirely when found; it only produces its own finding when the real upstream failure isn't otherwise visible as a `fail`-bucket check (`skills/fix-pr-review/SKILL.md:76`).

---
Created with LLM: Sonnet 5 | high | Harness: Claude Code

richkuo:
@claude sonnet review

richkuo:
Addressed review feedback (Sonnet 5 review bot · 2026-07-15T09:54:57Z) in e1972f3.

### Fixed
1. **Step 2's same-defect merge rule didn't extend to CI Failure findings** — confirmed: `SKILL.md:90` scoped the merge to "reviewer or thread" only, so a reviewer flagging the same defect a `fail`-bucket check already reports produced two findings instead of one, mirroring the gap already fixed intra-CI at line 76. Extended the merge rule to cover a defect raised by any source — reviewer, thread, or a CI Failure finding — noting all sources including the check name (`skills/fix-pr-review/SKILL.md:90`).

---
Created with LLM: Sonnet 5 | high | Harness: Claude Code

richkuo:
@claude sonnet review

richkuo:
Addressed review feedback (Sonnet 5 review bot · 2026-07-15T10:27:25Z) in 9100f77.

### Fixed
1. **Step 7's re-review routing narrowed CI blocking to only "fixed" failures, contradicting step 1's note** — confirmed: line 62 counts "any failing CI check" as blocking (verdict-independent, like the reviewer-section clauses beside it), but line 220 only counted "a fixed CI Failure." A CI failure wrongly refuted as flaky would downgrade to the lighter Sonnet re-review — exactly when the heavier Opus review is needed. Changed line 220 to count any CI Failure finding **regardless of verdict** (fixed or refuted), matching the reviewer-clause precedent and line 62, with the safety rationale stated inline (`skills/fix-pr-review/SKILL.md:220`). Chose this over narrowing line 62, since a refuted-but-real regression is precisely what the heavier review must catch.
2. **Cancel-bucket residual case had no concrete method to fetch upstream-failure detail** — confirmed: line 76 pointed at "wherever the run log points" while ruling out the only directly-reachable log. Added a concrete source — resolve the run ID from the cancelled check's `link`, then `gh run view <run-id>` lists sibling jobs in the same workflow run with their conclusions — plus an honest fallback: if no concrete failed job is findable, cite the cancellation with its run link and flag for human review rather than inventing a cause (`skills/fix-pr-review/SKILL.md:76`).

---
Created with LLM: Opus 4.8 | high | Harness: Claude Code

richkuo:
@claude review

PR #75, opened 2026-07-24
fix: this workflow uses `secrets: inherit` to pass a... in...

anupamme:
## Summary
Address high severity security finding in `templates/claude-workflow/workflows/claude.yml`.

## Vulnerability
| Field | Value |
|-------|-------|
| **ID** | yaml.github-actions.security.secrets-inherit.secrets-inherit |
| **Severity** | HIGH |
| **Scanner** | semgrep |
| **Rule** | `yaml.github-actions.security.secrets-inherit.secrets-inherit` |
| **File** | `templates/claude-workflow/workflows/claude.yml:300` |
| **Assessment** | Pattern match — needs manual review |

**Description**: This workflow uses `secrets: inherit` to pass all of the calling workflow's secrets to a reusable workflow. This violates the principle of least privilege because the called workflow receives access to every secret in the repository, not just the ones it needs. If the called workflow is compromised or sourced from a third party, an attacker gains access to all repository secrets. Instead, explicitly pass only the secrets that the called workflow requires using the `secrets:` map, e.g. `secrets: { MY_SECRET: ${{ secrets.MY_SECRET }} }`.

## Evidence

**Scanner confirmation**: semgrep rule `yaml.github-actions.security.secrets-inherit.secrets-inherit` matched this pattern as yaml.github-actions.security.secrets-inherit.secrets-inherit.

**Production code**: This file is in the production codebase, not test-only code.

## Threat Model Context

This is a Node.js library - vulnerabilities affect downstream consumers who use this package.

## Changes
- `templates/claude-workflow/workflows/claude.yml`
- `.github/workflows/claude.yml`
- `.github/workflows/claude-run.yml`

## Behavior Preservation
The change is scoped to 3 files on the vulnerable path, and the project's existing tests still pass, so intended behavior is unchanged.

## Verification
- [x] Build passes
- [x] Scanner re-scan confirms fix
- [x] LLM code review passed

---
*This change addresses a pattern flagged by static analysis. The code path handles user-influenced input and the fix reduces the attack surface against both manual and automated exploitation.*

---
*Automated security fix by [OrbisAI Security](https://orbisappsec.com)*

### DIFF
--- a/.github/workflows/claude-run.yml
+++ b/.github/workflows/claude-run.yml
@@ -81,6 +81,10 @@ on:
         required: false
         type: string
         default: ''
+    secrets:
+      CLAUDE_CODE_OAUTH_TOKEN:
+        description: 'Claude Code OAuth token for agent authentication'
+        required: true
 
 jobs:
   run:
--- a/.github/workflows/claude-run.yml
+++ b/.github/workflows/claude-run.yml
@@ -4,7 +4,9 @@ name: Claude Code run
 #
 # Published from richkuo/rk-skills (#5): consumer repos call this workflow via
 #   uses: richkuo/rk-skills/.github/workflows/claude-run.yml@main
-# with secrets: inherit. Prompt files and comment-patch scripts are fetched
+# passing CLAUDE_CODE_OAUTH_TOKEN explicitly under secrets: — the only secret
+# this workflow consumes, so callers need not (and should not) use
+# secrets: inherit. Prompt files and comment-patch scripts are fetched
 # from rk-skills at run time (templates/claude-workflow/), so updating
 # rk-skills updates every consumer repo on its next run with no per-repo sync.
 # A consumer repo tailors prompts two ways (both validated by the same
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -296,7 +296,8 @@ jobs:
       flow: ''
       model_id: ${{ needs.classify.outputs.model_id }}
       effort: ${{ needs.classify.outputs.effort }}
-    secrets: inherit
+    secrets:
+      CLAUDE_CODE_OAUTH_TOKEN: ${{ secrets.CLAUDE_CODE_OAUTH_TOKEN }}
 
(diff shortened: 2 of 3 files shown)


richkuo:
LGTM

Reviewed locally against the branch head: all three changed files inspected, both `claude.yml` copies confirmed still in sync (only the intended `runs_on: self-hosted` divergences), all workflow YAML parsed, and the full `bun test` suite run. CI `test` passes on the head commit.

The change is behavior-preserving. `CLAUDE_CODE_OAUTH_TOKEN` is the only caller-supplied secret `.github/workflows/claude-run.yml` consumes (lines 373 and 494); every other credential in that workflow is `github.token`, which is provisioned per-job and unaffected by the `secrets:` mapping. All three `secrets: inherit` call sites were converted, and the new `workflow_call.secrets` block is correctly placed as a sibling of `inputs`.

### Create Follow-up Issue

1. **Consumer repos still call this reusable workflow at a mutable `@main` ref.**
   `.github/workflows/claude-run.yml` is consumed cross-repo via `uses: richkuo/rk-skills/.github/workflows/claude-run.yml@main` (see `.github/workflows/claude.yml:293`, `:313`, `:351` and the same three call sites in `templates/claude-workflow/workflows/claude.yml`). Narrowing the secret surface from "every repository secret" to one token materially reduces blast radius, but the callee is still whatever currently sits on `main`: anyone able to push there changes what executes in every consumer repo on its next run, with that repo's token. Pinning callers to a release tag or commit SHA closes the remaining path. This is out of scope here — it is an independent versioning decision that affects every consumer repo and would require a release/tagging convention, so folding it into this diff would bloat and destabilize a focused security fix.
   **Plain simple English:** Other repos run this shared workflow by asking for "whatever is on the main branch right now." So any change pushed there runs everywhere immediately, using those repos' tokens. Pointing them at a fixed version instead would remove that risk. That is a separate piece of work, not part of this change.

---
Validated with LLM: Opus 5 | high | Harness: Claude Code

PR #76, opened 2026-07-24
feat(execution-block): make the fableplan stage effort stampable per issue [C34, Opus 5, high]

richkuo:
## Summary

The milestone pipeline hardcoded its fableplan planning agent to Fable 5 at `high` effort. An issue whose design is the whole risk got exactly the same planning budget as one that only needed its known work sequenced, and there was no way to say otherwise.

This adds an optional `- **Plan effort:** <low | medium | high | xhigh>` line to the `## Execution` block and plumbs it from prep through dispatch. Absent line means `high`, so every existing issue keeps its current behavior.

Design points worth flagging:

- **Effort only, never a model.** The fableplan stage is Fable 5 by definition, so the line sets effort and nothing else. `execution-plan-review` keeps the planner on Fable 5 if a revision tries to name one.
- **No tier is normalized.** Because the planner is always Fable, `low` and `medium` are legal here — unlike build effort, where they get raised to `high` on Opus/Sonnet/Haiku. This is the one place the Fable-only tiers pass through untouched.
- **The posted plan's footer now advertises the effort it actually ran at**, instead of always claiming `high`.
- **Inert stamps are refused, not written.** A `Plan effort` on a `fableplan: No` issue is never read, so `execution-plan-review` says so and asks rather than writing a value that does nothing.

## Verification

`bun test` — 76 pass, 0 fail (up from 75; 369 expect calls).

New coverage: dispatch at a stamped `xhigh` and `low`, the `high` default when the line is absent, no plan agent at all when `fableplan` is false, footer effort interpolation, and an assertion that nothing gets normalized.

One pre-existing test-harness gap surfaced while doing this: the shared pipeline harness had no `Plan` phase handler, so it threw `unexpected phase: Plan` and the two `fableplan: true` tests were silently exercising the planner-*failure* path while appearing to test planning. Added a default plan result, which moves them onto the success path they were meant to cover.

## Plain simple English

When the system plans a hard task before building it, that planning step always used the same amount of thinking. Now each task can ask for more thinking on its plan, or less, depending on how tricky the design is. Tasks that don't say anything keep working exactly as before.

---


richkuo:
@claude review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-07-25T04:11:07Z) in 618d11e. All four findings re-validated against current code and confirmed; none refuted. `bun test`: 81 pass, 0 fail.

### Fixed

1. **Hardcoded `high` in the fableplan plan footer** — the posted-comment template at `skills/fableplan/SKILL.md:78` is now `Created with LLM: <model that actually ran> | <effort that actually ran> | Harness: Claude Code | fableplan`, with prose forbidding a constant and pointing at the recorded values. Step 2 gained an explicit record-what-ran bullet (`skills/fableplan/SKILL.md:59`) so the footer resolves from what the dispatch actually did rather than from the stamped tier — covering all three must-survive cases: a `low` stamp reads `low`, an unstamped/unreferenced run names the session effort, and a fired model fallback names the substituted model alongside its real effort.

2. **`effort` passed to an Agent tool that may not accept it** — `skills/fableplan/SKILL.md:42` now does a pre-flight check on the Agent tool's own parameter schema in the running harness and omits the argument entirely when there is no `effort` property, plus a re-dispatch-once-without-it fallback if a call still fails input validation on it. Both paths are explicitly a degradation, never a step failure, and both feed the recorded effort into the footer. The schema check is the primary mechanism because it also closes the silent-drop case a reject-only fallback cannot detect: a harness that swallows the unknown parameter never errors, so the skill would otherwise report the stamped tier while planning at session effort. Confirmed the parameter's absence directly — this harness's Agent tool exposes only `description`, `isolation`, `model`, `prompt`, `run_in_background`, `subagent_type`.

3. **Inert `Plan effort` dropped in silence** (was Recommended Optional) — `workflows/milestone-pipeline.js:441-447` logs one line when a stamped `plan_effort` lands on a `fableplan: false` issue, alongside the existing stale build-effort and validate-effort logs. Worded as `ignoring Plan effort <tier> — fableplan is false, so no plan stage runs` rather than "normalized", since nothing is rewritten. Gated on `!missing_block` so the block-missing warning is not compounded by a second warning about a default it filled in, and gated on the field being present so an unstamped issue logs nothing. `tests/milestone-pipeline.test.js:285` now asserts the exact line for the existing `fableplan: false` + `Plan effort: xhigh` case, and a new test at `tests/milestone-pipeline.test.js:295` covers the two silent cases.

4. **Contract suite skipped the one file that consumes the field** (was Recommended Optional) — `tests/execution-block-contract.test.js:6` now reads `skills/fableplan/SKILL.md` alongside the other four, with four new tests: the footer template must carry placeholders and must not match `Created with LLM: Fable 5 | <tier> |` (so re-hardcoding fails), the `- **Plan effort:**` block line must appear in both `prd-to-issues` and `fableplan` and the field name in all four documents (so a one-sided rename fails on the mismatch rather than passing file-by-file), every document that states the default must state `high` together, and the effort-fallback instructions must be present. Verified these are real coverage and not tautologies: both defects above fail these assertions on the pre-fix file.

### Notes

No inline review threads were open, and CI was green (`test` check passing) before and after — this commit adds no new checks.

---

richkuo:
@claude review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-07-25T04:23:32Z) in e7af7ba. Both findings re-validated against current code and confirmed; neither refuted. `bun test`: 83 pass, 0 fail.

### Fixed

1. **Prep's fill-in default made `plan_effort` presence a meaningless signal** — confirmed. `plan_effort` is optional in `PREP_SCHEMA` (`workflows/milestone-pipeline.js:130` does not require it), but both the schema description and the prep prompt instructed a default of `high` when the line is absent, so a compliant prep agent returns the field on every issue and the log added in 618d11e would have fired for every `fableplan: false` issue carrying an Execution block. Prep is now contracted to **omit** the field when no `Plan effort` line exists (`workflows/milestone-pipeline.js:138`, `workflows/milestone-pipeline.js:420`), which is safe precisely because dispatch already applies the default at `workflows/milestone-pipeline.js:589` (`ex.plan_effort || 'high'`) — the remedy you proposed, and the right one. The comment at the log site (`workflows/milestone-pipeline.js:442-447`) now states the presence contract explicitly so the fill-in default is not reintroduced.

   Left `validate_effort` on its fill-in default deliberately: the runtime reads it directly and attaches no presence semantics to it, so the two fields are not inconsistent — only `plan_effort` needs absence to be meaningful.

   Coverage for all three must-survive cases: (a) the new `tests/milestone-pipeline.test.js:317` asserts against the prompt and schema source text that neither still says "default high when absent" — this is where the defect actually lived, so a harness-level test could not have caught it, since the harness supplies its own prep result; (b) `tests/milestone-pipeline.test.js:285` already asserts exactly one log naming `xhigh` for a genuine stamp; (c) `tests/milestone-pipeline.test.js:311-314` now also asserts an unstamped `fableplan: true` issue still dispatches at `high` with a matching footer, and logs nothing.

2. **Footer fallback named an unobservable session tier** (was Recommended Optional) — confirmed, and the point holds beyond style: an agent has no reliable read on its own effort tier, so "name the session effort" would have put a guessed value in the attribution slot this PR exists to make honest. `skills/fableplan/SKILL.md:57` and `skills/fableplan/SKILL.md:78` now fall back to the repo attribution default `high` when no tier was stamped or the harness could not honor one, stated explicitly as a documented default rather than an observation. Pinned at `tests/execution-block-contract.test.js:112`, including a negative assertion so reintroducing the session-effort wording fails. Case (c) is unchanged: the Notes model fallback still names the substituted model alongside whichever effort rule applied.

### Notes

Both new tests were checked against 618d11e and fail there, so they are real coverage rather than restatements of the fix. No inline review threads were open; the `test` check was green before and after.

---

richkuo:
@claude review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-07-25T04:32:04Z) in 6d6a3be. Verdict was `LGTM`; all three `Recommended Optional` items re-validated, confirmed, and implemented rather than deferred. `bun test`: 86 pass, 0 fail.

### Fixed

1. **A dropped `Plan effort` was never relayed to the operator who stamped it** — confirmed on both halves.

   *The notice:* `skills/fableplan/SKILL.md:86` now states the rule in step 5 — when a tier was stamped and the harness could not apply it, say so in one line, naming the tier and that the plan ran at the session's own effort instead. Both silent cases are spelled out so the notice cannot fire spuriously: an honored tier says nothing (must-survive b), and an unstamped or issue-less run makes no claim about a stamped tier in either direction (must-survive c), leaving the footer's `high` standing as the documented default. Step 2 now records *whether* the tier was honored, not just the value (`skills/fableplan/SKILL.md:57`), since that is what step 5 reads.

   *The explicit tier:* `skills/fableplan/SKILL.md:42` now passes `high` when nothing is stamped rather than omitting the parameter, so the footer's value is observed instead of conventional. Worth noting this is strictly better than I first credited: omitting does not merely leave the tier unknown, it can leave the plan running *below* `high` on a low-effort session while the footer claims `high` — the false-attribution direction this PR exists to close. The cost is that an `xhigh` session no longer implicitly plans at `xhigh`, which is the correct trade: `Plan effort: xhigh` is exactly the knob for that, and it is now honored wherever the harness allows it.

   This required updating step 1 (`skills/fableplan/SKILL.md:33`) and one existing assertion, which pinned the old "inherits the session effort" wording — the contract moved, so the test moved with it rather than being worked around.

2. **`execution-plan-review` created and retained inert stamps, and its table hid them** — confirmed on all three sub-parts, each fixed where the reviewer located it. `skills/execution-plan-review/SKILL.md:29` adds the strip-on-flip rule for a `Yes → No` revision, and extends it to any inert stamp already on an issue (offer to strip once, don't carry it forward). `skills/execution-plan-review/SKILL.md:46` scopes the missing-block backfill so a plan effort is derived only when the band puts fableplan at `Yes` — previously it stamped one on Capability 0/1/3 issues, writing the exact value the never-write rule forbids. `skills/execution-plan-review/SKILL.md:19` renders an existing inert stamp as `<tier> (ignored)`; the bare `—` is now reserved for issues that stamp no line at all, so a stale value is visible in the table rather than masked by it.

3. **The `missing_block` half of the log guard had no failing fixture** — confirmed, and self-inflicted: I removed `plan_effort` from that fixture in e7af7ba while addressing the previous round, which silently disarmed the assertion my own comment claimed to make. Restored at `tests/milestone-pipeline.test.js:303` with `plan_effort: 'xhigh'`, which is a realistic input precisely because it models a prep agent that fills the field despite the omit contract — the non-compliance the gate exists to absorb. Added a fourth fixture (`missing_block: true`, nothing stamped) so must-survive (c) is asserted explicitly rather than by implication. Verified by mutation: deleting `&& !normalized.missing_block` from `workflows/milestone-pipeline.js:446` now fails this test, where before it left the suite green.

### Notes

Every assertion added this round was checked against e7af7ba and fails there, so none of them merely restate the fix. No inline review threads were open; the `test` check was green before and after.

---

richkuo:
@claude sonnet review

PR #65, opened 2026-07-20
feat(#64): guard loop/validate shared pipeline rules against drift [C39, Cursor Grok 4.5, high]

richkuo:
## Summary

- Added `docs/contract-inventory.md` for the loop/validate skill families: review-cycle/LGTM stop (threshold 5), Capability gate (`< 2` / below 50 + safety carve-out), duplicate/convergence stop, and validation scope/feasibility/existing-PR stop — with owner/consumers/shared semantics/allowed divergence.
- Documented always-plan skills (`fable-validate-fableplan-loop`, `fableplan-loop`) as an intentional no-Capability-gate exception; documented `maxReviewCycles` workflow consumers as out of scope for this guard.
- Added `tests/loop-validate-pipeline-contract.test.js` semantic guard (key parameters, not exact strings); runs via existing `bun test` / CI.

Closes #64

## Verification

- `bun test tests/loop-validate-pipeline-contract.test.js` — 7 pass
- Red→green: temporarily changing `review_count > 5` → `> 99` in `work-on-issue-loop` failed the guard; restore passed
- `bun test` — 71 pass across 6 files

## Plain simple English

Shared “when to stop” rules for the loop skills now live in one checklist, and a test fails if any covered skill quietly drops those rules. Two skills that always plan on purpose stay allowed; the milestone workflow’s separate cycle setting is called out as not covered here.

---
Created with LLM: Cursor Grok 4.5 | high | Harness: Cursor

Issue #64, opened 2026-07-20
[C39] Dedupe the loop/validate skill families' shared pipeline rules


richkuo:
**Complexity: 39/100** — Capability 1 (Uncertainty 2 — deciding which wording divergences are intentional vs. a required shared contract needs author judgment); Volume 14 (Scope 3 — 11 skill files plus a new inventory doc and guard test; Coupling 2 — one shared prose-contract mechanism; Verification 2 — guard tests with per-rule fixtures) — Opus, high

## Problem

The `loop`/`validate` skill families restate the same pipeline rules as separate hand-written prose in each `SKILL.md`, with no single source of truth and no test that catches drift between copies:

- The review-cycle stop condition ("Stops on a bare LGTM with nothing left to fix; once past 5 review cycles it stops at the first LGTM it sees...") is written out near-verbatim in `skills/fix-pr-review-loop/SKILL.md:3` and `skills/work-on-issue-loop/SKILL.md:3`, and paraphrased again in `skills/fableplan-loop/SKILL.md:39`.
- The "skip fableplan when Capability < 2 / score < 50 with no safety flags" gate is restated in `skills/fable-validate-loop/SKILL.md:3` and `skills/validate-fableplan-loop/SKILL.md:3`. Two sibling skills state the inverse — always plan, no gate: `skills/fable-validate-fableplan-loop/SKILL.md:3` and `skills/fableplan-loop/SKILL.md:10`. That divergence is explicitly documented in both files ("no Capability gate — unlike fable-validate-loop") and is intentional — those skills exist to plan unconditionally — but nothing today records it as a deliberate exception or would catch a future copy that drifts silently.
- The "stopping instead when new-issue finds a duplicate or the discussion hasn't converged on one issue" condition is restated in `skills/new-issue-loop/SKILL.md:3` and `skills/fable-new-issue-loop/SKILL.md:3`.

This is the failure mode issue #57 generalized (repository rules drifting across hand-written copies) applied to the loop/validate families specifically — #57's own checklist named "Inventory loop and validation skill families and identify only the contracts that must remain identical" as a still-open item. #57 and its children (#58, #59, #60) were closed as superseded by this narrower issue, and PR #61 (the broader inventory + PR-review-format guard) was closed unmerged and is **not** a prerequisite or input to this issue — the inventory doc is created fresh here, scoped to the loop/validate families only.

## Goal

[...]

PR #77, opened 2026-07-25
feat(milestoneplan): add read-only milestone pre-flight before the workflow runs [C38, Opus 5, high]

richkuo:
## Summary

Nothing between `execution-plan-review` and `milestone-workflow` tells you, read-only, whether a milestone is actually runnable and what running it would cost. `execution-plan-review` renders and edits stamps issue by issue; `milestone-workflow` presents a run plan only after you've committed to running it.

`milestoneplan` is that pre-flight. It reads every issue in the milestone and returns:

- **Bucket classification** — build (open, no PR), resume (open with an open PR that closes it), skip (closed) — the same three buckets `milestone-workflow` uses, so the two skills agree on what a run would touch. GitHub's own `closedByPullRequestsReferences` linkage decides first (it catches a PR linked by hand with no closing keyword); the nine-keyword closing scan is the fallback.
- **The per-issue routing table** — one row per issue: state, bucket, complexity score, `Depends on`, `Runs after`, build model, effort, validate effort, fableplan, plan effort, and the first-review trigger. Missing fields are marked *missing* (never blank, never a guessed default). Only the runnable build-bucket rows claim to be what the run executes; closed and resume rows stay visible because a closed predecessor named in a runnable issue's `Depends on` has to be readable.
- **Band conformance** — recomputed from the `[C..]` score (`Capability = floor(score / 25)`, `Volume = score mod 25`). The band is a **floor**: an under-band build model is a finding, an over-band one is an observation with its own report section and no severity row, never a downgrade recommendation. The Volume tertile governs build effort only — validate effort and plan effort have their own rules, and a Fable build at `low` is the sanctioned discretionary tier.
- **Graph closure** — references that don't resolve, self-references, cycles across the union of both edge kinds, cross-milestone prerequisites, and edges into closed or resume predecessors. A closed predecessor's **merge state** is resolved, not assumed: batched aliased GraphQL bounded by the milestone (never a repo-wide merged-PR list), cross-repo lookups carrying `-R`, and timeline corroboration before an empty reference is read as "no PR".
- **Waves, critical path, and run size** — the shape of the run, the chain that gates the most work, and every term `milestone-workflow` step 2 computes, with the review-mode assumptions stated and each projected agent attributed to the model the pipeline actually dispatches it on (validate and plan are always Fable regardless of the build model; the first review defaults to Opus in subagent mode but is the build model in github mode).
- **A GO / GO WITH FINDINGS / NO-GO verdict**, with every finding class mapping to exactly one severity that matches what `milestone-workflow` would actually do — never blocking a run it would happily execute, never green-lighting an edge it would reject. A blocked subtree is excluded and the rest still runs.
- **Routing for each finding** — Execution-block fields go to `execution-plan-review`; score prefix and rationale-line items go to `validate-issue`, which owns body and title.

Severity is scoped to the buckets the run dispatches: a finding on a closed or resume issue is informational, because neither bucket's Execution block is ever read, and one pre-convention closed issue must not block every partially-completed re-run.

**Read-only by construction.** It never edits an issue body, posts a comment, opens a PR, or invokes the Workflow tool — guarded by a test rather than left to prose.

Wired in as stage 6 of `new-app-pipeline`, pushing `milestone-workflow` to stage 7. The stage is explicitly skippable.

## Verification

`bun test` — 125 pass, 0 fail.

The contract tests assert the skill's properties structurally rather than by grepping literals: the read-only guarantee (an open-world `gh` guard seeded with commands that must fail it, including the `--flag=value` write spellings), fetch-completeness discipline, bucket classification and the published closing-keyword pattern executed against all nine keywords, merge-state resolution and its cross-repo `-R` requirement, bucket-scoped severity and edge-finding ownership, one severity row per closed-predecessor edge kind, the model-attribution table, band derivation, the band table compared cell-for-cell against `validate-issue`'s canonical copy, the floor rule (over-band carries no severity row and does have a report section), run-size and review-worst-case parity with `milestone-workflow`, every severity having a section in the report template, and every finding class step 2 enumerates having a severity row — that last one derived from step 2's own flag lists, and negative-tested against an injected class with no row.

No install wiring needed — `bin/install.mjs` globs `skills/*/SKILL.md`. The skill runs inline, so no agent file either.

## Plain simple English

Before you set a batch of AI agents loose on a chunk of work, this checks the whole batch first. It shows every task with its settings, works out which tasks are already done or already in progress, catches broken or circular ordering between them, warns when a setting contradicts the task's own difficulty score, tells you the order things would run in and roughly how many agents it would take, and ends with a plain go or no-go. It only looks and reports — it never changes anything on its own.

---


richkuo:
@claude review

richkuo:
Addressed review feedback (@claude · 2026-07-25T04:11:25Z) in `fd8eee9`. Every finding re-validated against current code first; all six confirmed, none refuted. `bun test` — 88 pass, 0 fail.

### Fixed

1. **Subagent-mode review worst case carried into the projection** — step 4 now reproduces every term milestone-workflow step 2 computes, split by mode: `2×maxReviewCycles − 1` review-phase agents per issue as *direct* agents in the default `reviewMode: 'subagent'` (9 per issue at the default cap, so 90 not 10 on a 10-issue milestone), github mode's nesting exception where the worst case must **not** be applied, and `reviewLoop: false` zeroing the term (`skills/milestoneplan/SKILL.md:78-81`). The assumptions it projected under are now stated with the numbers (`:74`), the 1.5M projected-token `Large workflow` trigger is reported (`:83`), the resume bucket's `fix-pr-review-loop` cost is reported separately as outside the build-bucket sums (`:82`), and the retry term counts the runnable set rather than the milestone's full issue list (`:77`).

2. **Blocked subtrees are excluded and run around, not NO-GO'd** — verdict rules are now a table mapping every finding class to exactly one severity that matches what milestone-workflow actually does (`skills/milestoneplan/SKILL.md:117-136`). A hard edge into the resume bucket, a hard edge to an issue closed without a merged PR, and an open **hard** cross-milestone prerequisite are all *Blocked — excluded*: the dependent plus its hard descendants drop out and the rest runs, exactly as `skills/milestone-workflow/SKILL.md:19` does. NO-GO is reserved for cycles, missing Execution blocks, unresolvable references, a truncated fetch, and an empty runnable set — and an independent cycle still forces NO-GO alongside a merely-blocked subtree (`:136`). Open **ordering-only** cross-milestone prerequisites land at *Non-blocking* (unenforceable in `tracks`, so say so); a predecessor closed **with** a merged PR is a satisfied edge and no finding (`:56`). Exclusions propagate into the waves (`:64`), the projection (`:74`), the output template (`:97-113`), the hand-off (`:146`), and the failure-modes table.

3. **Band conformance is two-sided** — an over-band stamp is now a finding as silent overspend, with `workflows/milestone-pipeline.js`'s `model fable, effort high` default cited as the by-construction case; on-band Capability-3-at-Fable stays silent and an annotated departure stays a deliberate override (`skills/milestoneplan/SKILL.md:48`).

4. **The parity test now actually compares the two skills** — `tests/execution-block-contract.test.js:21-40` extracts each document's run-size formula (anchored on its sole `1 prep` code span) and the review worst case, and the tests compare the two extractions instead of pinning a literal in each. Seeded an extra term into milestone-workflow's formula: the test fails; reworded the prose around it: still passes. Also seeded a stale stage number: caught. The `/build.*resume.*skip/is` bucket test that asserted nothing is replaced with the three actual bucket definitions plus the closes-not-mentions rule (`:171-175`).

5. **Pipeline stage sequence made consistent, and enforced** — `README.md:76` now draws `execution-plan-review → milestoneplan → milestone-workflow`; `skills/milestone-workflow/SKILL.md:3` is Stage 7, `skills/execution-plan-review/SKILL.md:3` Stage 5, `skills/prd-to-issues/SKILL.md:3` Stage 4 (also stale by one, same defect class), and milestoneplan now declares Stage 6. A new contract test derives the stage table and asserts consecutive numbering, that every staged skill's self-declared number matches its table row, and that the README diagram walks the staged skills in table order (`tests/execution-block-contract.test.js:209-236`) — so a future mid-pipeline insertion cannot renumber one place and not the others.

6. **A truncated fetch can no longer produce a verdict** — `--limit` raised to 500, and the fetched count is reconciled against the milestone's own `open_issues + closed_issues` before anything is audited. A shortfall is a blocking unknown returning NO-GO, with the closed-heavy `--state all` case called out explicitly since that is where skip-bucket issues push build-bucket ones off the page (`skills/milestoneplan/SKILL.md:25`).

### Plain simple English

The cost estimate now shows the real worst case — a task can need nine review agents, not one — and says what settings it assumed. When one task is waiting on unmerged work, the tool no longer declares the whole batch unrunnable; it drops that task and prices the rest, matching what the actual runner does. It also flags needlessly expensive model choices, reads the whole milestone instead of the first hundred issues, and the tests now genuinely check the two documents agree.

---

richkuo:
@claude review

richkuo:
Addressed review feedback (@claude · 2026-07-25T04:30:03Z) in `a7e88b9`. All 7 findings re-validated against current code; all 7 confirmed, none refuted. `bun test` — 101 pass, 0 fail.

**Merge conflict resolved (the base branch moved to `e7af7ba` mid-review, leaving the PR `CONFLICTING`).** The only conflicted file was `tests/execution-block-contract.test.js`, and only its destructuring block — both sides had added readers. Reconciled by keeping both (`milestoneplan` + `new-app-pipeline` from this branch, `fableplan` from the base) plus the `validate-issue` reader the new band-table test needs; neither side's tests were dropped, and the base's 7 new Plan-effort tests pass unchanged. No other file conflicted. PR is `MERGEABLE` again.

### Fixed

1. **Every projected agent is now attributed to the model the pipeline actually dispatches it on** — step 4 gained a per-term model table (`skills/milestoneplan/SKILL.md:113-114` and the surrounding rows): validate **always Fable 5** (`workflows/milestone-pipeline.js:561`), plan **always Fable 5** (`:593`), first review the `PR review:` model **defaulting to Opus** (`:367`), review-loop fixers the Build model (`:397`), a re-review after a fix pass that cleared only non-blocking findings **Sonnet** (`:408`), and only implement following the Build model (`:609`). The worked example is stated: a 10-issue Capability-0 milestone is 10 Fable + 10 Sonnet + 10 Opus, not ~20 Sonnet. `reviewLoop: false` admits no reviewer/fixer model at all, and the subagent worst case's extra agents are split between the first-review model and the Build model.

2. **The milestone-size reconciliation is gone — completeness is proven from the fetch itself** (`skills/milestoneplan/SKILL.md:29`): a returned count strictly below its own `--limit` proves completeness; a count equal to the limit is indistinguishable from truncation and forces a re-fetch at a higher limit; only when no limit returns below itself is it a blocking unknown. Both of your suspected divergences are real and I verified them rather than assuming — on `rust-lang/rust`'s `1.0 beta` milestone the counters report 80 closed while `gh issue list` returns 79 and exactly one PR is assigned to it, and `GET /milestones` without `state` returned only the 2 open milestones out of 100. So the counters are now explicitly barred from triggering NO-GO (`:33`) and the milestones call passes `state=all` (`:21`) to keep a closed milestone auditable. This was a defect I introduced in the previous round, not a pre-existing one.

3. **Capability and Volume are derived from the score** (`skills/milestoneplan/SKILL.md:61`) — `validate-issue`'s `25 × Capability + Volume` with `Volume ≤ 24` makes `Capability = floor(score / 25)` and `Volume = score mod 25` fully determined by the `[C..]` prefix already parsed. A missing rationale line no longer suppresses the effort check, and a rationale Volume disagreeing with `score mod 25` is now its own finding.

4. **The read-only guarantee is tested as a property of the procedure, not its prose** (`tests/execution-block-contract.test.js:231`) — every `gh` invocation in the body's fenced blocks must be in a read-only allowlist, and no mutating `gh` verb may appear anywhere in the body (`:70`). On your must-survive list I deliberately took the stricter reading for the hand-off case: a mutating command has no place in this document even as an aside, because a reference can always be phrased by naming the sibling skill instead — and a missed mutation is a broken guarantee. `gh api` with `-X POST/PATCH/PUT/DELETE` or `-f`/`--field` also fails. Verified by seeding `gh issue edit` (fails) and `gh pr view` (passes).

5. **Bucket classification is one query** (`skills/milestoneplan/SKILL.md:41`) — `gh pr list --state open --limit 500 --json number,title,body,headRefName` once, then closing-keyword matching locally with a word boundary so `#12` never matches `#123`. Request count no longer grows with milestone size. A failed, throttled, or truncated lookup is a blocking unknown, never "no PR found" (`:51`) — that fall-through is exactly what would send an open PR back through a fresh build.

6. **The band table is compared against `validate-issue`, its canonical source** (`tests/execution-block-contract.test.js:311`, extractor at `:90`). The two tables carry different columns — `validate-issue` folds fableplan into the model cell and appends prose parentheticals, milestoneplan splits it out — so the extractor normalizes to comparable per-band tuples (capability, score band, model family, fableplan prescribed, three effort tertiles) and strips trailing parentheticals, which handles the row-3 discretionary-`low` phrasing you flagged as not comparable verbatim. Verified: shifting `25–49` to `25–48` in validate-issue fails the test; rewording the heading above the table does not.

7. **Body-content findings route to `validate-issue`, which can actually fix them** (`skills/milestoneplan/SKILL.md:185`, table row at `:202`). Step 6 now splits routing by finding class: Execution-block findings to `execution-plan-review` (its documented write scope, `skills/execution-plan-review/SKILL.md:35`), and body-content findings — missing acceptance criteria, missing problem statement, a missing or wrong `[C..]` prefix, a missing or contradictory rationale line — to `validate-issue`, which documents both `gh issue edit --title` and `--body-file` and owns the prefix and rationale line (`skills/validate-issue/SKILL.md:336`, `:338`). The "What this skill does not do" row that wrongly credited `execution-plan-review` with issue bodies is corrected.

### Plain simple English

The cost breakdown now names the right models: several steps always run on Fable or Opus no matter what each task says, so a batch that looked cheap was actually three different models. The completeness check I added last round was comparing two numbers that count different things — GitHub counts pull requests in a milestone, the issue list does not — so healthy batches were being called unrunnable; it now checks the download itself instead, and I confirmed the mismatch against a real repository. Finding in-progress work takes one lookup instead of hundreds, a failed lookup no longer looks like "nothing in progress", and findings about a task's written content now point at the tool that can actually edit it.

---

richkuo:
@claude review

richkuo:
Addressed review feedback (@claude · 2026-07-25T04:47:56Z) in `da28e10`. All 5 findings re-validated against current code; all 5 confirmed, none refuted. `bun test` — 105 pass, 0 fail. PR was already `MERGEABLE`, CI green — no conflict work this round.

### Fixed

1. **Closed-predecessor merge state is now fetched, so the distinction the severity table turns on is decidable** — you were right that nothing step 1 queried could tell a satisfied edge from an unsatisfiable one. Step 1's issue fetch now carries `stateReason,closedByPullRequestsReferences` (`skills/milestoneplan/SKILL.md:22`), and since that field names the closing PRs but **not** whether they merged (verified against `cli/cli` — the nodes return `number`/`repository`/`url`, no merge state), each **distinct** closing PR is resolved once with `gh pr view <n> --json number,state,mergedAt` (`:64`). Bounded by the number of closing PRs on closed predecessors, not by milestone size, so it does not reintroduce the per-issue growth the last round removed. Merge state decides, never `stateReason` (`:72`) — a `NOT_PLANNED` issue whose closing PR merged is satisfied, a `COMPLETED` one with no merged PR is not. An unresolvable lookup is an *indeterminate* edge reported as a blocking unknown (`:74`), with its own NO-GO row (`:188`), rather than silently taking the blocking branch — which on a re-run would have excluded every merged-predecessor subtree. Both new commands were already inside the read-only allowlist, so the guarantee is unchanged.

2. **All nine closing keywords, case-insensitive** (`skills/milestoneplan/SKILL.md:49`, pattern at `:54`) — `close|closes|closed|fix|fixes|fixed|resolve|resolves|resolved`, with the optional `owner/repo` prefix for the cross-repo form and the trailing boundary that keeps `#12` from matching `#123`. The contract test no longer greps for keyword names: it extracts the published pattern and **executes** it (`tests/execution-block-contract.test.js:267`) against all nine keywords in both lowercase and capitalized form, `fixes owner/repo#12`, a bare `see #12` mention, `reverts #12`, and `closes #123`. Verified by narrowing the pattern back to three keywords — the test fails on the first missed keyword.

3. **Both resume-bucket edge kinds have a disposition, and the cost carries its sequencing** (`skills/milestoneplan/SKILL.md:101`, severity row at `:190`) — a hard edge stays *Blocked — excluded*; an ordering-only edge is **satisfied by sequencing**, because `milestone-workflow` step 1 runs that PR's `fix-pr-review-loop` to completion before invoking the pipeline, so it is reported as informational with the gating PR named rather than going unmentioned. Step 4's resume-bucket line now states that cost is sequenced *ahead of* the build bucket, not concurrent with it (`:129`) — the number was previously readable as overlapping.

4. **The prep row names a model** (`skills/milestoneplan/SKILL.md:137`) — the pipeline passes no `model`, only `effort: 'low'`, so prep inherits the session model (Opus 5 in this repo's normal usage; a Fable session makes it a Fable agent). Every attribution row now places into the per-model mix, prep included. A new test asserts every row of that table names a model rather than a cost adjective (`tests/execution-block-contract.test.js:317`).

5. **Duplicated hand-off lead-in removed** — step 6 carried "Offer exactly the next actions the verdict supports…" both before and after the routing rules. The routing block keeps its own lead-in (`skills/milestoneplan/SKILL.md:208`) and the offer list keeps exactly one.

### Plain simple English

When a task depended on earlier work that was already finished and merged, the tool had no way to see that it merged, so it assumed the worst and could call a healthy batch unrunnable — it now looks the merge up, and says "I don't know" instead of guessing when the lookup fails. It also recognizes all nine phrases GitHub accepts for closing an issue instead of three, so work written up as "Fixed #12" is no longer planned twice. In-progress work is now correctly shown as running before the main batch rather than alongside it, and the cost table names the model for its first step instead of just calling it cheap.

---

richkuo:
@claude review

richkuo:
Addressed review feedback (@claude · 2026-07-25T05:02:25Z) in `bc6848f`. All 5 findings re-validated against current code; all 5 confirmed, none refuted. `bun test` — 113 pass, 0 fail. PR is `MERGEABLE`/`CLEAN` against `main`, CI green — no conflict work this round.

### Fixed

1. **Severity now derives from the bucket the finding's issue sits in, so a closed issue can no longer NO-GO a milestone the runner would finish** — confirmed the whole chain you traced. `milestone-workflow` step 1 drops closed issues from the plan and runs resume issues through `fix-pr-review-loop` outside the pipeline, and the pipeline's prep agent reads only the issues that arrive in `tracks` (`workflows/milestone-pipeline.js` builds its list from the tracks it was handed, then prep reads exactly that list), so the old row's stated reason — "prep routes it on conservative defaults" — was false for anything but the build bucket. The audit still runs milestone-wide (`skills/milestoneplan/SKILL.md:82`), because bucket membership is not decidable without the whole set and a closed issue is still a predecessor whose merge state a runnable edge turns on; only *severity* scopes. Runnable issues keep the table's severity (`:199`), and one precedence row demotes any otherwise-blocking finding whose issue sits in the skip or resume bucket to informational (`:211`) rather than patching row by row — that keeps the "exactly one severity" invariant intact instead of duplicating each row per bucket. A resume finding is reported as *deferred, not resolved*, naming the re-entry: if that PR closes unmerged the issue returns to the build bucket and the finding is blocking again. The step 1 fetch is the one milestone-wide exception, since an incomplete fetch is precisely what makes buckets undecidable. Same reasoning applied to the cycle row, which had the identical flaw: a cycle routed through a skip- or resume-bucket node never enters `tracks`, so it is a filing defect, not a NO-GO (`:198`).

2. **The review term is attributed per mode, and github mode names the Build model** (`skills/milestoneplan/SKILL.md:156`) — verified at `workflows/milestone-pipeline.js:655`: in `reviewMode: 'github'` the single `review-loop` agent is dispatched with `model: modelId` and `effort: ex.effort`, both from the issue's Execution block, and `reviewLoopPrompt` never references `first_review_model` — so the `PR review:` line is genuinely never read on that path. The table row says so explicitly, step 4's github-mode bullet now names that model instead of leaving the term unattributed, and the worked example gives both mixes side by side (`:158`): the same 10-issue Capability-0 milestone is 10 Opus reviewers in subagent mode and **no Opus at all** in github mode. Also stated that the reviews those loop agents respond to come from the `@claude` Action, which is not a pipeline agent and so sits outside the projected counts entirely — otherwise the github-mode number reads as if the review itself were free.

3. **The merge lookup targets the closing PR's own repository** (`skills/milestoneplan/SKILL.md:64`, rule at `:67`, failure mode at `:268`) — you were right, and this one was mine from the last round. The command now passes `-R <owner>/<repo>` taken from the reference's own `repository`; verified the shape against a real closing reference in this repo, which returns `repository: {name, owner: {login}}`, so `repository.owner.login` / `repository.name` are both already in hand from the step 1 issue fetch — no extra request. Called out why this is worse than an error: a bare `gh pr view 42` returns a confident verdict in *either* direction, satisfied or unsatisfiable, from an unrelated PR. An unreadable cross-repo PR falls to the existing indeterminate rule as a blocking unknown.

4. **The read-only guard is open-world** (`tests/execution-block-contract.test.js:98`) — confirmed: the extractor matched seven hardcoded subcommands, so `gh release create` or `gh workflow run` matched neither it nor `MUTATING_GH` and passed unseen. It now matches any `gh <subcommand>` and fails on anything not provably read-only, checks `--method` alongside `-X`, and scans prose as well as fenced blocks for the `gh api` write forms. Rather than assert inline, the logic is extracted into `readOnlyViolations()` returning violation strings, which makes the guard testable in both directions (`:295`): seeded `gh release create` / `gh workflow run` / `gh label create` / `gh secret set` / `gh api --method POST` / `-X PATCH` / `-f` all produce violations, and every command this procedure actually uses produces none. One thing the open-world match surfaced: `\b` is load-bearing in the extraction regex — without it the trailing `gh` of "through" matches and the next prose word reads as a subcommand. That's commented at the call site.

5. **Each closed-predecessor edge kind has exactly one severity row** (`skills/milestoneplan/SKILL.md:206`, prose at `:115`) — the unqualified duplicate is gone; the hard-edge row keeps *Blocked — excluded* and the ordering-only row is *no finding*, since an ordering edge only prevents overlapping work and a closed issue is dropped from the plan, leaving nothing to overlap. Both rows now name the closed-with-no-PR / closed-unmerged case in their own text rather than referring to "that same predecessor" positionally, so the pairing survives a reordering of the table. The step 2c prose carries the same rule, and the invariant line now reads "none of them falls through the rules, and none of them lands in two".

Every one of the five was checked by seeding the failure it claims to catch and confirming it fires: unqualifying the Execution-block NO-GO row, dropping `-R`, reverting the github-mode row to Opus, re-adding the duplicate severity row, and adding `gh release create` to the procedure each fail exactly the intended test and nothing else.

### Plain simple English

The tool was calling a batch of work unrunnable because of problems in tasks the real runner never even looks at — already-finished ones, and ones already being reviewed — so a single old leftover task could block a batch that would run fine. Severity is now judged only on the work that will actually be dispatched, and problems elsewhere are reported as notes. The cost table also said the review step always runs on one fixed expensive model; in one of the two review setups it runs on whatever model the task is built with, so both setups now show their own real cost. Two smaller fixes: when earlier work was finished in a different code repository, the tool looked the number up in the wrong repository and could confidently read the wrong answer; and a test meant to prove this step never changes anything on GitHub only inspected seven kinds of command, so an unrecognized new one would have slipped through.

---

richkuo:
@claude review

richkuo:
Addressed review feedback (@claude · 2026-07-25T05:23:30Z) in `519198d`. 6 findings re-validated against current code: 5 confirmed, 1 partial. `bun test` — 119 pass, 0 fail. PR is `MERGEABLE`/`CLEAN`, CI green — no conflict work this round.

### Fixed

1. **The finding taxonomy actually closes now** (`skills/milestoneplan/SKILL.md:206` for the section, `:213` for the mapping, `:232`–`:239` for the new rows) — confirmed both halves. The template had *Blocking*, *Blocked — excluded*, *Non-blocking*, *Deliberate overrides* and no home for **Informational**, which is the severity the bucket-scoping rule from `bc6848f` emits in three rows — so every demoted finding vanished, including the resume-bucket one step 2 explicitly requires be reported rather than dropped. Added the section plus a stated severity→section mapping (a NO-GO row prints under *Blocking*; a *no finding* row prints nowhere). On the second half, every class you listed did lack a row; each now has one derived from what `milestone-workflow` actually does with it rather than from taste: a self-reference is **NO-GO** (a degenerate cycle — no edge from an issue to itself can be ordered), missing ordering fields whose *kind* the prose does not establish is **NO-GO** (step 1 refuses to guess hard-vs-ordering and the two produce different waves), and inferable edges, missing acceptance criteria or problem statement, a missing `[C<score>]` prefix, a predecessor in both fields, and a duplicate within a list are all **Non-blocking** — the runner executes all of them. The stale-model-name case folded into the existing "stamp that lies about what will run" row instead of getting its own, since a duplicate row is the defect the last round removed; the pipeline falls back to `fable`, so the class is identical. The two classes that genuinely resolve through other rows now say so (`:242`) instead of reading as omissions.

2. **Merge state is corroborated, and the empty-array inference is gone** (`skills/milestoneplan/SKILL.md:86`) — see the partial note below on the mechanism, but the invariant holds and is implemented: an empty array is no longer a verdict. When it is empty, the merged-PR list already fetched is scanned first — a merged PR that closes the issue by keyword makes the edge satisfied (the reference was simply never recorded), a merged PR that merely *mentions* it also makes it satisfied and is reported as informational with the PR named (the one place a bare mention carries weight, since the code is in the base branch either way), a **complete** merged list with no reference *establishes* "closed with no PR" so the hard-edge exclusion stands, and a truncated or failed list is indeterminate — a blocking unknown. That covers all four of your must-survive cases, including the genuinely PR-less issue staying excluded rather than becoming a permanent unknown.

3. **One merged-PR query instead of one lookup per predecessor** (`skills/milestoneplan/SKILL.md:66`) — you're right that the old bound was illusory exactly where it mattered: on a partially-completed re-run nearly every issue is closed and most are predecessors, so "distinct closing PRs" tracks the milestone's size. Now `gh pr list --state merged --limit 500 --json number,mergedAt,body` once, matched locally, under the same truncation rule as the issue fetch. Cross-repo references keep their own targeted `-R` lookup, since a list of *this* repo's merged PRs cannot contain them — that count is bounded by cross-repo references, which does not grow with the milestone.

4. **GitHub's own linkage is the primary bucket signal, keyword regex the fallback** (`skills/milestoneplan/SKILL.md:49`) — confirmed with a live counterexample rather than on principle: `cli/cli` open issue `#13968` is linked to open PR `#13969` whose body carries no closing keyword for it, so the keyword-only scan would have put an issue that already has a PR into **build** and opened a duplicate. `closedByPullRequestsReferences` reports it, costs no extra request, and covers both linkage forms. Also fixed a trap I had left in the published pattern: the inner `(es|ed)` was a capturing group, so an implementer's "group 2 is the issue number" would have read `es` — it is now `(?:es|ed)`, and the doc says group 2 is always the number.

5. **The milestones call paginates** (`skills/milestoneplan/SKILL.md:21`, rule at `:25`) — measured on `rust-lang/rust`: the bare call returns 30, `per_page=100` returns 100, `--paginate` returns all 140. So on a repo with more than 30 milestones the "list them and ask which" step showed a silent subset and a named milestone could read as not found. Now `--paginate` with `per_page=100`, and the prose says why this fetch gets no exemption from the step's completeness discipline.

6. **The read-only guard catches every write form of `gh api`** (`tests/execution-block-contract.test.js:92`) — `-F`/`--raw-field` and `--input` POST exactly as surely as `-f`, and both are now violations. One thing your suggestion would have broken if applied literally: `gh api graphql` *always* POSTs and always carries `-f`/`-F` parameters, so flagging the parameter forms unconditionally would fail a legitimate read-only GraphQL query — the kind this repo's own tooling uses. GraphQL is therefore judged by operation keyword instead: a `query` passes, a `mutation` is a violation. Both directions are seeded in the guard's own test.

### Corrected scope (partial)

1. **Empty-reference finding — the invariant is real, the suspected mechanism is not.** You flagged `includeClosedPrs` defaulting to false as the concrete risk and asked me to confirm it; I checked, and it does not hold. Issue `#64` in this repo is closed, its closing PR `#65` is `MERGED`, and `gh issue list --json closedByPullRequestsReferences` returns that PR — so merged closing PRs are not hidden from `gh`, and satisfied edges on a re-run do not come back empty. The fix `da28e10` landed is not inverted. What *is* real is the narrower case you listed under must-survive: an issue closed by hand after its work merged under a PR that only mentions it has no reference at all, so absence still cannot be read as "no PR". Fixed on that basis, with the verified behavior recorded in the doc (`:86`) so the next reader does not re-litigate it.

Every one of the six was checked by seeding the failure it claims to catch — deleting the Informational section, reverting the empty-reference inference, restoring the per-predecessor lookup, dropping the sidebar-linkage rule, unpaginating the milestones call, removing a severity row, and adding a new severity to the table with no matching section each fail exactly the intended test and nothing else.

### Plain simple English

Two real problems fixed. The report the tool prints had no place for one whole category of note, so those notes disappeared; and several kinds of problem it looks for — a task missing its acceptance criteria, or its score — were never assigned a seriousness. Both are now complete, with a test that fails if a future category is added without somewhere to print it. It also no longer treats silence from GitHub as proof that earlier work had no pull request; it double-checks against the list of merged work first. Smaller fixes: it asks once about merged work instead of once per finished task, it now notices pull requests attached by clicking rather than by writing "closes #12", and it reads past the first thirty milestones. One claim did not survive checking: the worry that GitHub hides already-merged pull requests turned out to be wrong, and the check is recorded so nobody redoes it.

---

richkuo:
@claude review

richkuo:
Addressed review feedback (@claude · 2026-07-25T05:42:46Z) in `0710e0a`. All 4 findings re-validated against current code; all 4 confirmed, none refuted. `bun test` — 120 pass, 0 fail. PR remains `MERGEABLE`/`CLEAN`; the `test` check is green.

### Fixed

1. **Bucket-demotion row scoped to findings the skip/resume issue owns** — step 2 now defines edge-finding ownership ("An edge finding has two endpoints, and its owner is the endpoint this run would dispatch" — `skills/milestoneplan/SKILL.md:113`), and the demotion row explicitly exempts cross-bucket edge rows, whose owner is the runnable dependent (`skills/milestoneplan/SKILL.md:252`). All four must-survive cases hold and are pinned by a new contract test (`tests/execution-block-contract.test.js:446`): hard edge into resume stays *Blocked — excluded*, a closed issue's missing Execution block still demotes, a resume issue's band finding demotes with the re-entry named, and a build-bucket–internal hard edge has no finding for the demotion to touch.
2. **Merge-state fetch bounded by the milestone, not repo history** — the repo-wide `gh pr list --state merged` (bodies included, limit-escalated, NO-GO on unresolvable truncation) is gone. Step 1 now collects the distinct closing PR numbers on closed predecessors that build-bucket issues hard-depend on and resolves them in one batched aliased GraphQL query per ~50 (`skills/milestoneplan/SKILL.md:63`), keeping the cross-repo `-R` lookup. The empty-`closedByPullRequestsReferences` corroboration reads the issue's own `CROSS_REFERENCED_EVENT` timeline (bounded per issue, batchable, keyword check runs over the returned bodies) instead of the repo-wide list (`skills/milestoneplan/SKILL.md:88`). A genuinely unresolvable lookup stays a blocking unknown. Both new GraphQL commands were verified live against this repo before being published in the doc.
3. **Rationale-line classes given severity rows; coverage test derives its class list** — two new Non-blocking rows: missing complexity rationale line, and a published Volume contradicting `score mod 25` (`skills/milestoneplan/SKILL.md:247`), both routed to `validate-issue` as step 6 already prescribed. The coverage test now derives step 2's finding classes from its flag lists and stem-matches each against the severity table (`tests/execution-block-contract.test.js:659`) — negative-tested: injecting a fabricated rowless class fails it.
4. **Stale-model reasoning corrected** — the example is now a genuinely unmappable name (non-Anthropic), with the doc stating that any Opus spelling maps to `opus` per `workflows/milestone-pipeline.js:417` (`skills/milestoneplan/SKILL.md:117`); the severity rationale now states the real consequence — prep's output schema (`workflows/milestone-pipeline.js:119`) coerces an unmappable name to one of the four model ids unpredictably, and the `fable` default applies only to a missing Execution block (`skills/milestoneplan/SKILL.md:251`).

---

richkuo:
@claude review

richkuo:
Scope change per user direction, addressing the review of 2026-07-25T05:42:46Z's successor (the 06:04:41Z review) in `9c99fc4`: the skill is rescoped to only the per-issue routing table (score, both dependency fields, build model, effort, validate effort, fableplan, plan effort, first review) plus band-derived recommendations for changing those fields. Bucket classification, merge-state resolution, graph checks, waves, run-size projection, and the go/no-go verdict are removed. `bun test` — 101 pass, 0 fail. PR body updated to match.

### Fixed

1. **Placeholder form** — the one surviving fetch (the milestones lookup) now uses the documented `{owner}`/`{repo}` form (`skills/milestoneplan/SKILL.md:21`), and the read-only guard fixture matches (`tests/execution-block-contract.test.js:319`). For the record: the `:owner` colon form inside `gh api graphql -F` values was verified working live in this repo before the rescope, so this is consistency, not a functional fix.

### Resolved by removal

1. **Openness of `closedByPullRequestsReferences` entries** — bucket classification is gone; the skill no longer classifies build/resume/skip at all.
2. **Cursor argument on the cross-reference query** — the merge-state resolution and its timeline corroboration are gone.

---

richkuo:
@claude review

richkuo:
Addressed review feedback (@claude · 2026-07-25T06:31:46Z) in `741e7ae`. All 4 findings re-validated against current code; all 4 confirmed, none refuted. `bun test` — 103 pass, 0 fail. PR remains `MERGEABLE`/`CLEAN`.

### Fixed

1. **Scope table/"what the run will execute" and recommendations to open issues** — only open rows are claimed as dispatchable; closed rows stay visible for context (incl. closed predecessors) but are marked not dispatched; recommendations are open-only, with closed contradictions listed as informational (`skills/milestoneplan/SKILL.md:54`, `:60`–`:67`, `:86`–`:90`). Resume caveat for open-issue+open-PR noted in one line (`:54`).
2. **Volume-tertile check scoped to build effort; Fable `low` exempted** — tertile applies to **build** effort only; Validate/Plan judged by their own rules; Fable discretionary `low` never recommended against (`skills/milestoneplan/SKILL.md:48`).
3. **Truncated issue fetch disclosed** — count equal to `--limit` triggers re-fetch or an explicit partial-table statement; under-limit milestones print no caveat (`skills/milestoneplan/SKILL.md:27`, `:92`).
4. **`--flag=value` forms caught by the read-only guard** — `WRITING_GH_API` matches `=`-attached method/field/input spellings; seeded `--method=POST`, `-X=PATCH`, `--field=title=x` (and `-f=`/`-F=`/`--raw-field=`/`--input=`) must fail (`tests/execution-block-contract.test.js:71`–`:76`, `:316`–`:322`).

---
Created with LLM: Grok 4.5 | high | Harness: Cursor

richkuo:
@claude review

richkuo:
Addressed review feedback (@claude · 2026-07-25T06:45:50Z) in `0e49fee`. All 3 findings re-validated against current code; all 3 confirmed, none refuted. `bun test` — 104 pass, 0 fail. PR remains `MERGEABLE`/`CLEAN`.

### Fixed

1. **Band is a floor — no over-band downgrade recommendations** — under-band still recommends a raise; over-band is an *Over-band observation*, never a change recommendation to weaken the stamp; safety carve-out sentence copied verbatim from `validate-issue` with an explicit never-downgrade rule for money/data/security/auto-protective bodies (`skills/milestoneplan/SKILL.md:46`, `:69`). Band-parity test now compares the floor + carve-out sentences too (`tests/execution-block-contract.test.js`).
2. **Resume caveat is unconditional and qualifies recommendations** — always stated (relevance undecidable without the removed PR lookup); every open-issue recommendation line carries the resume uncertainty (`skills/milestoneplan/SKILL.md:56`, `:62`).
3. **Validate effort applies its documented rules** — vocabulary `medium|high`, default high, `medium` on-rule only for Capability 0 with Volume ≤ 7, `low` flagged as out-of-vocabulary silent coercion, `xhigh` still banned (`skills/milestoneplan/SKILL.md:48`).

---
Created with LLM: Grok 4.5 | high | Harness: Cursor

richkuo:
@claude review

richkuo:
Scope reversal per user direction, in `51f3146`: the capabilities cut by the earlier rescope (`9c99fc4`) are restored, and the corrections made after it are carried forward rather than reverted with it.

### Restored

Fetch-completeness discipline · build/resume/skip bucket classification · the nine-keyword closing scan behind GitHub's own `closedByPullRequestsReferences` linkage · merge-state resolution of closed predecessors (milestone-bounded batched GraphQL, cross-repo lookups carrying `-R`, timeline corroboration on an empty reference) · graph closure with per-edge-kind dispositions · execution waves and critical path · the run-size projection and its per-model attribution table · the severity table · the GO / GO WITH FINDINGS / NO-GO verdict · body auditing beyond the score prefix and rationale line.

### Carried forward, not reverted

The `{owner}`/`{repo}` placeholder form · band-as-floor (under-band is a finding; over-band is an observation with its own report section and no severity row, never a downgrade recommendation) · the Volume tertile scoped to build effort only · the full Validate effort rules (vocabulary, default, the Capability-0 medium case, the silent `low`→`medium` coercion) · the sanctioned Fable discretionary `low` · the `--flag=value` gh api write guard.

### Reconciled

Two post-rescope fixes were phrased around capabilities that no longer existed, and are now expressed through the restored machinery rather than dropped:

1. The unconditional resume caveat existed because the rescoped skill had no PR lookup and could not tell when a resume applied. With bucket classification back it can, so the caveat is replaced by the actual bucket, which is strictly more informative — the test now asserts the bucket scoping instead of the caveat text.
2. Open-only claim scoping becomes runnable-build-bucket scoping. The per-issue table gained `State` and `Bucket` columns and covers every issue in the milestone, so a closed predecessor named in a runnable issue's `Depends on` stays readable while only the runnable rows claim to be what the run executes.

`bun test` — 125 pass, 0 fail. The derived severity-row coverage check was negative-tested against an injected finding class with no row, and fails as intended.

---

richkuo:
@claude review

richkuo:
Addressed review feedback (@claude · 2026-07-25T07:15:53Z) in `ed267ee`. All 5 findings re-validated against current code; all 5 confirmed, none refuted. `bun test` — 126 pass, 0 fail. PR remains `MERGEABLE`/`CLEAN`.

### Fixed

1. **"Open" for `closedByPullRequestsReferences` is established by intersecting with the open-PR list** — same-repo reference numbers must appear in the `:44` open-PR list to classify **resume**; a same-repo reference absent from that list is not resume; a cross-repo reference whose openness cannot be established is a **blocking unknown**, never assumed open or closed (`skills/milestoneplan/SKILL.md:51`).
2. **Build-effort tertile check is one-sided like the model check** — under-tertile is a finding; over-tertile is an *Over-band observation* (section widened to model or build effort), never a downgrade recommendation; safety carve-outs cover effort too (`skills/milestoneplan/SKILL.md:133`–`:135`, `:223`, `:257`, `:262`).
3. **Closing-keyword `owner/repo` prefix is captured and compared** — bare `#12` and same-repo `owner/repo#12` count; foreign `otherorg/other#12` is discarded (`skills/milestoneplan/SKILL.md:59`–`:61`).
4. **Cross-reference query takes `$after`** — `timelineItems(..., after:$after, ...)` with per-issue paging instructions (`skills/milestoneplan/SKILL.md:91`–`:97`).
5. **"Silently" corrected for logged normalizations** — non-Fable `low`/`medium`→`high` and Validate `xhigh`→`high` are logged; Validate `low` remains the unlogged out-of-schema coercion, with undetermined allowed-tier pick stated (`skills/milestoneplan/SKILL.md:135`, `:257`).

---
Created with LLM: Grok 4.5 | high | Harness: Cursor

richkuo:
@claude review

richkuo:
Addressed review feedback (@claude · 2026-07-25T09:39:03Z) in `c4abc16`. All 4 findings re-validated against current code; all 4 confirmed, none refuted. `bun test` — 128 pass, 0 fail. PR remains `MERGEABLE`/`CLEAN`.

### Fixed

1. **Step-1 bucket-classification blocking unknowns map to NO-GO** — severity rows for a failed/throttled/truncated open-PR query and an undecidable linked-reference openness; NO-GO trigger list now covers every step-1 blocking unknown (`skills/milestoneplan/SKILL.md` severity table + verdict rules).
2. **Cross-repo linked PR openness resolved via existing `gh pr view -R` lookup** — readable repos resolve to open/not-open; unreadable or errored lookups stay blocking unknown → NO-GO (`skills/milestoneplan/SKILL.md:51`).
3. **Validate-effort `low` reason corrected** — prep schema maps `low→medium` (not an undetermined pick); no-runtime-log clause kept (`skills/milestoneplan/SKILL.md:135`, severity row).
4. **One placeholder spelling** — GraphQL `-F` values use `{owner}`/`{repo}` to match the REST form and `fix-pr-review`'s documented auto-fill convention.

---
Created with LLM: Grok 4.5 | high | Harness: Cursor

richkuo:
@claude review

richkuo:
Addressed review feedback (@claude · 2026-07-25T09:50:24Z) in `4e3ed93`. All 4 findings re-validated against current code; all 4 confirmed, none refuted. `bun test` — 131 pass, 0 fail. PR remains `MERGEABLE`/`CLEAN`.

### Fixed

1. **Out-of-milestone referenced issues are fetched** — bounded `gh issue view` / aliased GraphQL for number/state/stateReason/milestone; absence from the milestone set is never scored as nonexistent; failed lookup → NO-GO (`skills/milestoneplan/SKILL.md:39`–`:47`, severity row).
2. **`fableplan: No` on Cap-2 exempt when build is already Fable 5** — planning inherent; Cap-2 + Opus + No stays a finding (`skills/milestoneplan/SKILL.md:143`).
3. **Stamp recommendations cover excluded build-bucket issues** — they re-enter once the blocker clears; skip/resume stay informational (`skills/milestoneplan/SKILL.md` hand-off).
4. **Closed predecessor with open closing PR has severity rows** — hard → Blocked — excluded; ordering-only → no finding; not routed through resume (`skills/milestoneplan/SKILL.md:94`, severity table).

---
Created with LLM: Grok 4.5 | high | Harness: Cursor

richkuo:
@claude review

richkuo:
## Disposition — review after `4e3ed93`

| # | Finding | Verdict | Action |
|---|---|---|---|
| 1 | Out-of-milestone lookup must include `body` + `closedByPullRequestsReferences` and recurse until the graph closes | ✅ Confirmed | Fetch now asks for those fields and repeats until no new numbers appear, bounded by closed dependency closure size |
| 2 | Fable `low` exemption only for Volume ≤ 7 (`[C75]`–`[C82]`) | ✅ Confirmed | Discretionary tier scoped; higher-Volume Fable `low` stays under-tertile |
| 3 | Add Per-model mix line to report template | ✅ Confirmed | Template line added after Run size |
| 4 | Mark excluded build rows as `build (excluded)` | ✅ Confirmed | Bucket cell distinguishes runnable vs excluded |

Pushed in `94deab7`. Contract tests updated; `bun test` green (133).

---
Updated with LLM: Grok 4.5 | high | Harness: Cursor

richkuo:
@claude review

richkuo:
## Disposition — review after `94deab7`

| # | Finding | Verdict | Action |
|---|---|---|---|
| 1 | Missing ordering field with no implied edge was NO-GO | ✅ Confirmed | Split into three rows: ambiguous kind stays NO-GO; inferable stays Non-blocking; prose implies no edge is Non-blocking (satisfied empty graph / stamp `none`) |
| 2 | Excluded build-bucket findings had no severity | ✅ Confirmed | Excluded NO-GO-class → Informational (deferred); Non-blocking stamps stay Non-blocking; NO-GO verdict never from excluded-only |
| 3 | Concurrency + token threshold missing from template | ✅ Confirmed | Added `Concurrency:` line and `token Large-workflow at 1.5M` in Run size assumes |
| 4 | Flag-list split cut mid-parenthetical | ✅ Confirmed | Top-level semicolon split + exact derived class count (24) |
| 5 | Band table not compared to `prd-to-issues` | ✅ Confirmed | Third-copy parity against validate-issue |

Pushed in `$(git rev-parse --short HEAD)`. `bun test` green (136).

---
Updated with LLM: Grok 4.5 | high | Harness: Cursor

richkuo:
@claude review

richkuo:
## Disposition — review after `1722535`

| # | Finding | Verdict | Action |
|---|---|---|---|
| 1 | Cycles through out-of-milestone nodes / ordering-only into excluded had no severity | ✅ Confirmed | Widened Informational cycle row to any non-dispatched node (incl. out-of-milestone); ordering-only into excluded = *no finding*; hard stays under descendant exclusion |
| 2 | Template `<worst>` undefined in github / reviewLoop:false | ✅ Confirmed | Slot defined per mode: subagent worst case / github (= planned) / review off (= planned) |
| 3 | Token-target / budgetFloor missing from projection | ✅ Confirmed | Added Token-target deferral bullet + template line; parity claim includes that term |
| 4 | execution-plan-review Fable `low` unscoped | ✅ Confirmed | Allowed only at Volume ≤ 7 (`[C75]`–`[C82]`), matching milestoneplan |

Pushed in `ecb7722`. `bun test` green (140).

---
Updated with LLM: Grok 4.5 | high | Harness: Cursor

richkuo:
@claude review

richkuo:
## Disposition — review after `ecb7722`

| # | Finding | Verdict | Action |
|---|---|---|---|
| 1 | Closed cross-milestone edges reported as satisfied / no finding | ✅ Confirmed | In-milestone merged → still *no finding*; closed hard out-of-milestone → Blocked — excluded; closed ordering-only → Non-blocking |
| 2 | Per-model mix unbound vs three run-size figures | ✅ Confirmed | Template labels mix with `<bound>`; retries → Fable 5 |
| 3 | Ambiguous missing-field kind was NO-GO | ✅ Confirmed | Non-blocking — same as workflow's plan-review flag |
| 4 | Demotion excepted only the fetch row | ✅ Confirmed | All four step-1 blocking unknowns exempt from bucket demotion |

Pushed in `330217c`. `bun test` green.

---
Updated with LLM: Grok 4.5 | high | Harness: Cursor

richkuo:
@claude review

richkuo:
## Disposition — review after `330217c`

| # | Finding | Verdict | Action |
|---|---|---|---|
| 1 | Closed hard cross-milestone (merged) was Blocked — emptied v1+ runs | ✅ Confirmed | Merged → Non-blocking stamp rewrite to `none`; unmerged → still Blocked — excluded |
| 2 | Ambiguous-kind edges had no kind for hard-only rules | ✅ Confirmed | Treat as ordering-only until stamped; say so on waves / critical path / exclusion |
| 3 | Out-of-milestone predecessors missing from table | ✅ Confirmed | Context rows with Bucket `external` |
| 4 | Read-only guard missed line-continued `gh api` writes | ✅ Confirmed | Join `\\\n` before scanning; seeded two-line cases |
| 5 | Resume findings had no recommendation path | ✅ Confirmed | Resume joins runnable/excluded for stamp recommendations; skip stays note-only |

Pushed in `7f1d94b`. `bun test` green (146).

---
Updated with LLM: Grok 4.5 | high | Harness: Cursor

richkuo:
@claude review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-07-25T10:50:07Z) in `b1d2741`.

### Fixed
1. **Resume stamp findings had no printable recommendation slot** — resume stamp/band findings keep Non-blocking (with re-entry note); only NO-GO-class demotes to Informational (`skills/milestoneplan/SKILL.md:122`, demotion row, failure mode).
2. **Fable `low` exemption was Volume-only** — scoped to Capability 3 ∧ Volume ≤ 7 (`[C75]`–`[C82]`) in milestoneplan + execution-plan-review; Cap-0/1/2 at `low` stays a finding.
3. **Read-only guard missed non-`gh` writes** — deny-list for `git commit/push/…` and mutating `curl`; seeded cases + read passthroughs.

`bun test` green (146). No blocking findings in this review set.

---
Created with LLM: Grok 4.5 | high | Harness: Cursor

richkuo:
@claude sonnet review

PR #92, opened 2026-07-28
refactor(milestoneplan): remove the Recommended fix column from the findings table [C4, Fable 5, high]

richkuo:
## Summary
- Removes the `Recommended fix` column from milestoneplan's findings table, per repo-owner direction: the audit is display-only. The table is now `Severity | # | Finding | Route` — the `Finding` cell names the defect, `Route` names the owning skill, and the corrected value is derived by that skill from the same score formula, never printed by the audit.
- Deletes the column's cell-value vocabulary (`add — …` / `none — …` fixed forms) and the per-cell band/tertile derivation requirement; updates step 6's routing paragraph accordingly.
- This moots the last open review finding on PR 83 (undefined cell values for audit-tooling `Blocking` rows) by construction — there is no cell left to define.
- This commit was pushed to the PR 83 branch after its squash-merge, so it lands here as its own PR (cherry-picked, no content changes).

## Verification
- `bun test` — 157 pass, 0 fail on top of current `main`.
- Tests pin the 4-column header/separator shape, drop the removed-column contract checks, and guard against a `Recommended fix` column returning.

## Plain simple English
The milestone pre-flight report had a column that suggested exact corrected settings for each problem it found. We removed it: the report now just names each problem and which tool fixes it, and that tool computes the right value itself — so the suggestion and the rulebook can never disagree.

---


richkuo:
@claude review

PR #100, opened 2026-08-01
feat(milestone-pipeline): batch github review cycles [C55, GPT-5, high, fableplan]

richkuo:
## Summary

- Keep GitHub review cycle 1 with the implementation agent that opened the pull request.
- Give each later fix agent up to two review cycles before the workflow starts a fresh agent.
- Validate each agent's reported cycle count and preserve the existing review stopping rules and merge contract.
- Update milestone run-size calculations for the lower agent count.

## Verification

- `bun test` — 127 pass, 0 fail.

## Plain simple English

The agent that builds the change now handles the first review round. Each later helper can handle two rounds. This reduces helper startup work while still rotating helpers before their memory grows through many review rounds.

---
Updated with LLM: GPT-5 | high | Harness: Codex


richkuo:
@claude review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-08-02T01:34:05Z) in dc12415.

### Fixed

1. **Merge gate trusts the self-reported cycle-1 verdict** — GitHub mode now makes the independent merge agent verify the newest completed bot review, exact `LGTM` verdict, trigger order, run success, and unchanged readiness head immediately before the pinned merge; a changed head requires a fresh review (`workflows/milestone-pipeline.js:481`).

Verification: `bun test` (128 pass).

---
Created with LLM: GPT-5 | high | Harness: agent

richkuo:
@claude sonnet review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-08-02T02:07:16Z) in d9f7f70.

### Resolved judgment calls (was Requires Human Review)

1. **Confirm whether `issue_comment` runs report the default-branch SHA before trusting the merge gate's skip of a head-sha cross-check** — **Confirmed; the skip is correct and stays.** Two independent sources: GitHub's Actions event reference lists `issue_comment` as `GITHUB_SHA` = "Last commit on default branch" / `GITHUB_REF` = "Default branch"; and this repo's own `@claude sonnet review` run [30728273779](https://github.com/richkuo/rk-skills/actions/runs/30728273779) reports `event: issue_comment`, `headBranch: main`, `headSha: 80d70d8` while the PR head at that moment was `dc12415`. `.github/workflows/claude.yml:2-3` confirms the trigger is `issue_comment`, so every `@claude review` run takes this path. A `head_sha` cross-check against the PR would therefore reject every legitimate review.

   **Went further than the recommended solution.** Keeping the instruction as-is would have left the gate correct but incomplete: with the SHA cross-check validly unavailable, the standing review was tied to the merged code only through the review loop's *self-reported* readiness SHA — the same self-report class the previous cycle's finding was about. The ordering rules bind the LGTM to the trigger comment, not to the commit, so a push that lands after an LGTM (agent pushes a straggler fix, then reports the new head) would pass the gate with a review of older code. The gate now adds an independent, server-clocked time binding: resolve when the readiness head became visible on GitHub (earliest `created_at` among that SHA's check-suites, falling back to the commit's committer date when it has none) and require the LGTM output's `created_at` to be strictly later; an LGTM that predates its head stops the merge (`workflows/milestone-pipeline.js:483`). Covered by `tests/milestone-pipeline.test.js:1024-1046`.

   Alternatives rejected: restore the `head_sha` cross-check — disproven above, it would block every review; parse a SHA out of the `@claude` comment body — the Action posts a branch link, not a commit SHA, so there is nothing to bind to; require a fresh review unconditionally before every merge — correct but it re-runs a review on unchanged code every time and does not close the gap any better than the timestamp check. The check fails safe: clock skew can only block a merge and surface a blocker, never pass a stale review. Override if you would rather keep the gate exactly as reviewed.

**Verification:** `bun test` — 128 pass, 0 fail. CI `test` check green on the prior head.

---

richkuo:
@claude review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-08-02T02:20:15Z) in f0b03b4.

### Fixed

1. **The batch-review prompt offers "1 or 2" cycles even on a batch that owns only one cycle** — Confirmed and fixed at both sites. The return line now derives its gloss from `cycleLimit` ("exactly 1" at a one-cycle batch, "1 or 2" otherwise), matching the pluralization line 370 already did (`workflows/milestone-pipeline.js:383`). Went one step past the recommended wording change on the schema: `GITHUB_REVIEW_BATCH_SCHEMA` was a module constant that could not see `cycleLimit`, so re-wording alone would have left it a static gloss. It is now `githubReviewBatchSchema(cycleLimit)` with `minimum: 1, maximum: cycleLimit` (`workflows/milestone-pipeline.js:234-240`, applied at `415`). An out-of-range `cycles_run` now fails structured-output validation and the agent retries, instead of returning cleanly and tripping the bounds check at `422` — so the failure mode the finding describes cannot reach the loop at all, not merely be discouraged by wording.

   All three "must survive" cases are covered by a new parametrized test (`tests/milestone-pipeline.test.js:795-808`): `maxReviewCycles: 2` (first batch owns one cycle → "exactly 1", `maximum: 1`) and the default (batch owns two → "1 or 2", `maximum: 2`), with an explicit assertion that the self-contradictory "1 or 2, never above 1" string can no longer be rendered. The `maxReviewCycles: 4` final-batch case takes the identical `cycleLimit === 1` path as the first case. The existing bounds check at `422` stays as the defense in depth behind schema validation.

**Verification:** `bun test` — 130 pass, 0 fail (128 before, +2 new).

**On the review's verification limitation:** the GitHub Actions events reference was unreachable on that run. It was confirmed on the prior cycle from two sources — the docs page itself (`issue_comment` → `GITHUB_SHA` "Last commit on default branch") and this repo's own run [30728273779](https://github.com/richkuo/rk-skills/actions/runs/30728273779) (`event: issue_comment`, `headSha: 80d70d8` = main, PR head `dc12415`). As this review notes, that claim is not load-bearing for merge safety either way.

---

richkuo:
@claude sonnet review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-08-02T02:47:07Z) against f0b03b4 (no code change).

### Not changed (refuted)

1. **Review cycle 1 can spend two real `@claude` review rounds while the pipeline charges one cycle** — A cycle is one review-resolution pass over a standing verdict, not one bot request. The build pass and later passes use the same read, resolve, re-trigger, and wait sequence (`workflows/milestone-pipeline.js:348`, `workflows/milestone-pipeline.js:370`). The owner skill defines cycle 1 this way and defines `maxReviewCycles` as a stopping rule, not a bot-request cap (`skills/milestone-workflow/SKILL.md:37`, `skills/milestone-workflow/SKILL.md:61`).

---
Created with LLM: GPT-5 | high | Harness: agent

richkuo:
@claude sonnet review

PR #102, opened 2026-08-03
feat: validate issues under C20 on Opus 5 at medium [C34, Opus 5, medium]

richkuo:
## Summary

Validation in the milestone pipeline used to run on Fable 5 for every issue, at the stamped `Validate effort`. Below a `[C20]` score an issue is Capability 0 with small Volume, where a whole-issue Fable pass buys nothing a cheaper reviewer misses.

The pipeline now derives the validation model from the `[C<score>]` prefix:

| Score | Validate model | Validate effort |
|---|---|---|
| `[C1]`–`[C19]` | Opus 5 | medium (any stamped tier goes unread, and the run logs that) |
| `[C20]` and above | Fable 5 | the stamped `Validate effort`, default high |
| no `[C..]` prefix | Fable 5 | default high — unknown is not small |

The routing is **derived, never stamped**. No Execution block gains a `Validate model:` line, so no issue can carry a stale value that nothing reads — the same reason `Plan effort` is refused on `fableplan: No` issues.

## Verification

- `bun test` — 131 pass, 0 fail.
- New test `routes validation off the score` covers all five cases above, including the exclusive boundary at exactly C20 and the log line emitted when a stamped effort is skipped.
- The milestoneplan contract test that asserted the old always-Fable claim now asserts the score-derived rule instead. It was not weakened or removed; its premise was the policy this PR changes.

## Docs updated

`milestoneplan`, `execution-plan-review`, `milestone-workflow`, and `prd-to-issues` all state the new rule. `execution-plan-review` also gains a failure-mode row rejecting any attempt to stamp a validate model or a sub-C20 validate effort.

## Plain simple English

Before a job is built, an agent checks that the job description is correct. That check always used the most expensive model. For very small jobs, this PR uses a cheaper model at lower effort instead. The size score on the job decides, so nobody has to set it by hand.

---
Created with LLM: Opus 5 | medium | Harness: Claude Code


richkuo:
@claude review

PR #103, opened 2026-08-04
feat(fable-validate-fableplan): add plan-only counterpart to fable-validate-fableplan-loop [C6, Opus 5, high]

richkuo:
## Summary

Adds `fable-validate-fableplan` — `fable-validate-fableplan-loop` with the implementation stage removed.

The chain head is identical: `fable-validate` runs, the same scope gate stops the run on a too-large / architecturally-infeasible / already-addressed issue, the update-issue edits land when the verdict calls for them, and `fableplan` always posts a Fable 5 plan (no Capability gate). The run then stops at the posted plan and names the follow-on options instead of starting one. No worktree, no build, no PR, no review loop.

The always-plan rule is deliberate here for a second reason beyond the loop variant's: the plan is this skill's only output, so gating it away would leave the run with nothing to deliver.

## Verification

- `bun test` — 130 pass, 0 fail.
- The new skill is registered in `tests/loop-validate-pipeline-contract.test.js` under both `ALWAYS_PLAN` (must state the missing Capability gate) and `VALIDATION_STOP` (must carry the three STOP rows), so the shared guard covers it.

## Changes

- `skills/fable-validate-fableplan/SKILL.md` — new skill.
- `tests/loop-validate-pipeline-contract.test.js` — added to the two guard consumer lists; one test title widened from "loops" to "chains".
- `docs/contract-inventory.md` — inventoried file row, always-plan and validation-stop consumer lists, and the intentional-exception note.
- `README.md` — Fable-driven skills table row.

## Plain simple English

There is already a skill that checks a GitHub issue with Fable, corrects it, plans it, then builds it. This adds the same thing without the build. It checks the issue, fixes the description, posts a plan, and stops. Use it when a person must decide what to do with the plan.

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,7 @@ Every issue's first line also carries an explicit **`fableplan: yes|no`** signal
 | `fable-validate` | Like `validate-issue`, but the fact-checking runs on a Fable 5 subagent; your main session presents the verdict and acts on it. |
 | `fable-validate-loop` | Runs `fable-validate`, applies issue fixes, gets a Fable plan (only when Capability ≥ 2 / score ≥ 50, or touching safety-critical code), then drives to a reviewed PR. |
 | `fable-validate-fableplan-loop` | Same as above, but the Fable plan is unconditional — every issue gets a posted plan before implementation, no matter how simple. |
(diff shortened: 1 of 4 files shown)


richkuo:
@claude review

PR #110, opened 2026-08-07
test(contracts): guard verdict template and report cap in loop skills [C4, Fable 5, medium]

richkuo:
## Summary
- Add two semantic guards to `tests/loop-validate-pipeline-contract.test.js`: (1) the verdict-block template that 5 loop skills quote and parse must keep every parsed field co-located on one line, matched against the canonical format in `validate-issue`; (2) the "Cap the whole report at 55 words, ELI18" rule must stay present in all 9 autonomous chain skills.
- Record both rules as new rows in `docs/contract-inventory.md` with owner, consumers, required semantics, and allowed divergence.

## Design note
The original proposal was to extract these blocks into shared reference files. That conflicts with the repo's contract design: the existing tests require stop semantics in each skill's procedure body, so an autonomous loop carries its own stop conditions instead of depending on an optional file read. The inventory also states there is no generator for SKILL.md prose. This PR therefore guards the copies rather than moving them.

## Verification
- `bun test` — 132 pass, 0 fail.
- Mutation check: changing "55 words" to "80 words" in `fableplan-loop` and deleting the `fableplan:` field from `validate-issue-loop`'s template both fail the new tests; restoring the files returns the suite to green.

## Plain simple English
Nine automation skills repeat two important text blocks. If one copy changes and the others do not, the automation can break. This change adds tests that compare all copies. When a copy drifts, the tests fail and show the file.

---

https://claude.ai/code/session_01KM6HNnzLmDhbvsBt6ZUhZv


richkuo:
@claude sonnet review

PR #119, opened 2026-08-08
feat(#116): escalate the review band from the measured diff [C37, Opus 5, high]

### LINKED ISSUE #116: [C37] Review routing does not escalate when the diff is much larger than the scored volume
**Complexity: 37/100** — Capability 1 (Uncertainty 2 — diff-to-volume mapping and thresholds are undesigned); Volume 12 — Opus 5, high · fableplan: no

## Problem

The six-band matrix routes the first review from the issue's `[C<score>]` prefix. The score is a prediction made at issue time. The diff is a measurement made at build time. When a low-scored issue produces a very large diff, the routed reviewer is too weak for what actually shipped. Example: a C15 issue that produces a 900-line diff still gets the bare `@claude` band-1 review.

## Goal

At review-trigger time, compare the actual diff to the scored Volume. When the diff implies a higher band, escalate the review to that band's review model. Escalation is upward only. A small diff must never lower the review tier, because small diffs can still carry Capability-3 risk (money, data integrity, security).

## Approach

1. Derive an implied Volume from the diff (changed lines and file count, mapped to the existing 0–24 Volume scale). Exclude noise: lockfiles, generated files, and pure renames.
2. Compute an implied score with the title's Capability unchanged: `25 × Capability + impliedVolume`.
3. If the implied score lands in a higher band than the title score, trigger the higher band's review model and effort. Record a `rescore`-style entry, mirroring the existing validation escalation.
4. Surface the divergence to the orchestrator so the `[C..]` prefix can be restamped.

This corrects Volume under-scoring only. It cannot detect under-scored risk; the safety carve-out still covers that.

## Acceptance criteria

- The pipeline escalates the review band when the implied score crosses a band boundary upward.
- The pipeline never lowers the review band on a small diff.
- Thresholds and the diff-to-Volume mapping are documented in `validate-issue` (or the pipeline doc) with the exclusion list.
- Tests cover: escalation across a boundary, no-op when bands match, no downgrade, and exclusion of lockfile/generated churn.

---

richkuo:
Closes #116

## Summary

The `[C..]` score routes the first review. That score is a prediction made when the issue is filed; the diff is a measurement made when the pull request is open. A low-scored issue that produced a very large diff still drew the reviewer its score routed.

At review time the pipeline now measures the pull request's **effective diff** and applies two independent rules. Either can raise the review band; neither can lower it.

### Rule 1 — implied score from the canonical Volume axes

The diff re-derives the three real axes, and the scale's own formula does the rest:

`Volume = (Scope + Coupling + Verification) × 2`, `implied score = 25 × Capability + Volume`

Capability stays the title's (`floor(score / 25)`), because no diff evidences Risk or Uncertainty.

- **Scope** — file count and line count, the larger step of the two.
- **Coupling** — distinct top-level directories (subsystem proxy) and distinct file extensions (language proxy), the larger step. Named as proxies: a diff shows how far a change spreads, never whether the parts talk, so this axis reads low by design.
- **Verification** — test files and test lines the diff carries, the larger step.

Full step tables are in `validate-issue` under **Diff-measured review escalation**.

### Rule 2 — the size floor

Faithful axes alone barely move a Capability-0 issue: leaving band 1 needs Volume 21 of 24, so nearly every axis has to max out. A diff past the point where one reviewing pass can hold it still needs a capable reviewer, whatever the axes say.

So an effective diff of **600 or more changed lines, or 25 or more changed files**, lifts the review to **band 2 (Opus 5 · high)** on its own. Both thresholds are judgment calls, checked against this repository's 40 most recent merged pull requests so a routine change never trips them.

The two rules are recorded separately. Only rule 1 claims the filed score was low, so only rule 1 offers a `[C..]` restamp; the floor moves the reviewer without disputing the score.

### The effective diff

Excluded from every measurement: lockfiles by basename, generated output by path segment and suffix, and every file that moves no line (pure renames, mode changes, binary swaps). The list is deliberately short — excluding a file only shrinks the measured surface, so a wrong entry weakens the review instead of strengthening it.

### Rules both obey

- **Upward only.** A small diff never lowers the review tier: it can still carry Capability-3 risk (money, data integrity, security).
- **Neither demotes a stronger reviewer.** A stamped `PR review:` trigger that already outranks the escalated band stands; the divergence is still recorded.
- **Only the review moves.** Build model, build effort, and fableplan are untouched, because the build already ran.
- **Neither detects under-scored risk.** The safety carve-out still covers that.

**Default github review mode.** Cycle 1's trigger is composed before the diff exists, so an escalated pull request is re-reviewed on the escalated trigger even when cycle 1 already returned `LGTM`. When cycle 1 needs updates, the escalated trigger rides on the batch's normal re-trigger instead of spending a cycle. A run that exhausts `maxReviewCycles` before the escalated review lands ends `max_cycles_exhausted` rather than approved.

**Surfacing.** Each escalated issue carries a `review_escalation` record in the run results — implied score, Volume, the three axes, band, which rule fired, reviewer, measured diff, and whether the reviewer actually changed. `milestone-workflow` step 5 reports it and offers the restamp only when rule 1 fired.

Measurement and policy are split on purpose: the implementation agent returns raw per-file diff stats (`diff_files`, now required in `IMPLEMENT_SCHEMA`), and the runtime owns the exclusions, the axis tables, and the floor, so the policy is testable and no arithmetic runs inside an agent.

### Note on `bandFor`

`bandFor(0)` means "no `[C..]` prefix — unknown, route as the top band". A computed implied score of 0 is genuinely zero, so the implied score goes through a new `bandForScore`, the plain range lookup. Reusing `bandFor` here would have escalated every tiny diff to band 5.

## Verification

- `bun test` — 178 pass, 0 fail (was 160 before this change; 18 new tests).
- New tests cover the acceptance criteria: escalation across a band boundary, a no-op when the bands match, no downgrade on a small diff, and exclusion of lockfile, generated, and pure-rename churn.
- Also covered: each rule firing on its own with its recorded axes and `reason`, the restamp offered for rule 1 and withheld for rule 2, a case just under both floor thresholds holding its reviewer, a stamped stronger reviewer surviving an escalation, an unmeasured diff reported instead of silently skipped, both github-mode paths (bare `LGTM` and `needs updates`), and cycle exhaustion refusing to approve.
- Red → green proven: the behavior tests failed against the unmodified pipeline, and temporarily disabling the escalation guard re-broke the three github-mode guarantees before the guard was restored.

## Plain simple English

The system chose the code reviewer from a score written when the task was filed. That score can be too small for the code the task actually produced. Now the system measures the real change when the pull request opens. If the change spans more surface than the score claimed, or is simply too large for one reading pass, it moves the review to a stronger reviewer. It never moves to a weaker one.

---


richkuo:
Closing without merging.

This PR implemented what issue #116 asked for: the pull request diff measured against the filed `[C..]` score at review time, with the review band raised when the diff implies a bigger job. On review of the scope, that mechanism is not wanted.

The behavior the repository does want — models and efforts re-routed when validation returns a higher complexity — already ships on `main` at `workflows/milestone-pipeline.js:811-824`. A validated score in a higher band re-routes build model, build effort, fableplan, and the first review, drops the stale stamps, and records a `rescore` so the orchestrating session restamps the issue.

No change is needed, so nothing here merges. The branch `cc/issue-116-review-routing-diff-escalation` stays available if the diff-measured routing is ever wanted.

---
Validated with LLM: Opus 5 | high | Harness: Claude Code

PR #121, opened 2026-08-08
refactor(skills): run sync-docs, create-release, and sync-docs-release inline [C28, Opus 5, high]

richkuo:
## Summary

`sync-docs` and `create-release` were dispatch shims: each one immediately handed its work to a runner subagent that held the real workflow. `sync-docs-release` chained both, so a docs-and-release run placed two subagents between the user and every change.

This removes the indirection. Each workflow body now lives in its own `SKILL.md`, both runner agents are deleted, and `sync-docs-release` invokes the two skills in-session. The procedures themselves are unchanged — only where they run moves.

- `skills/sync-docs/SKILL.md` and `skills/create-release/SKILL.md` carry the full workflows verbatim.
- `agents/sync-docs-runner.md` and `agents/create-release-runner.md` are deleted; `agents/` is now empty and dropped from the published package.
- `skills/sync-docs-release/SKILL.md` states that no step delegates to a subagent.
- `bin/install.mjs` and `install.sh` remove the two retired runner files from `~/.claude/agents`, so an earlier install cannot dispatch to a stale copy.

## Verification

`bun test tests/` — 160 pass, 0 fail. The drift guard in `tests/sync-docs-release-contract.test.js` now reads the two skill files instead of the two agent files, and every shared-semantics assertion still holds against the moved text.

## Tradeoff

Subagents kept the two long procedures out of the main session's context. Running inline spends that context in exchange for visible edits and the ability to ask a question mid-run.

## Plain simple English

Two skills did no work themselves. They sent the job to a helper agent, and you saw only a short report. Now each skill does the work in your session, so you see every change as it happens. The combined docs-and-release skill also runs fully in your session.

---


richkuo:
@claude review

PR #125, opened 2026-08-10
docs(claude-md): name the claim/uncertainty separation in the litotes rule [C5, Opus 5, high]

richkuo:
## Summary

The "No litotes" rule in the Response Style section already required the confidence statement to sit in its own clause or sentence. That instruction appeared after the list of banned phrasings, so it read as one more example of the ban. This change states the separation principle as a named rule, placed as the second sentence of the bullet, before the examples.

Added sentence: `Keep the claim and the uncertainty in separate clauses.`

A named principle covers phrasings the example list does not enumerate.

## Verification

- `CLAUDE.md` line 19 and `AGENTS.md` line 19 both carry the new sentence.
- Both files stay byte-identical to each other, as they were before this change.
- Documentation only. No code, tests, or skill logic changed.

## Plain simple English

The writing rule told me to state my confidence in a separate clause. That instruction was easy to miss, because it came after a list of banned phrases. This change puts the rule in its own sentence near the start. Now the principle is clear for all cases. It applies past the listed examples.

---


richkuo:
@claude sonnet review

PR #128, opened 2026-08-11
feat(install): install skills for Codex and Cursor alongside Claude Code [C39, Opus 5, high]

richkuo:
## Summary

`npx rk-skills` and `install.sh` both wrote into `~/.claude` only, so Codex and Cursor users got nothing. All three agents auto-discover the same `SKILL.md` folder format, so the skills themselves already worked there — the installers and the docs were the part that assumed one harness.

**Installers**

- `bin/install.mjs` installs into every harness home present on the machine (`~/.claude`, `~/.codex`, `~/.cursor`). New flags: `--claude` / `--codex` / `--cursor`, `--all`, `--help`; `--project` still switches the scope root.
- Cursor also scans the Claude Code and Codex skill roots, so auto-detection skips Cursor's own root when one of those is installed. Without that, every skill would appear in Cursor two or three times. `--cursor` forces it.
- Two correctness fixes fall out of the rewrite: the installer no longer merges into an existing skill directory (which left files from a removed older version behind), and it no longer copies over a symlink pointing into a local checkout, which would have written the package's own files back through the link.
- `install.sh` links into the Claude Code and Codex roots, links `AGENTS.md` into `~/.codex/`, and prunes the duplicate rk-skills symlinks earlier runs left in Cursor's two skill directories. The prune only unlinks symlinks whose target resolves into this repo; anything else in those directories is untouched.

**Skills**

- Fable-model skills and the two Workflow-tool skills now open with a `> **Harness:** Claude Code only` line naming what they need and what to run instead. A Codex or Cursor user now finds out at the door rather than partway through a run.
- `work-on-issue` and `fableplan` told Cursor and Codex to create their worktrees inside `.claude/` — another agent's directory — and disagreed with each other about the path. Both now use the repo's harness-neutral `.worktrees/`, and `.gitignore` covers it plus the per-agent worktree directories.
- `EnterWorktree` was documented as using the passed name verbatim. It does not: `cc/multi-harness-install` produced branch `worktree-cc+multi-harness-install` in this very session, so a later `git -C` anchored on the reconstructed path would miss. `work-on-issue`, `fableplan`, `CLAUDE.md`, and `AGENTS.md` now say to read the real path from the tool result and the real branch from `git branch --show-current`.
- `app-prd` and `sync-docs-release` hardcoded the `cc/` branch prefix; both now name the per-agent prefix.

## Verification

- `bun test` — 185 pass, 0 fail (163 before; `tests/multi-harness.test.js` adds 22).
- Installer exercised against scratch directories in `--project` scope: fresh directory, Codex-plus-Cursor (dedup path), Cursor alone, `--all`, and a re-install over a live symlink. The symlink survived and the source checkout stayed clean.
- Harness paths were confirmed against the installed tools, not from memory: the Codex binary states it installs skills into `$CODEX_HOME/skills/<skill-name>` (default `~/.codex/skills`), and Cursor's own `skill-path-utils` lists `.cursor/skills/`, `.claude/skills/`, `.codex/skills/`, and `.agents/skills/` among its skill roots.

## Plain simple English

The skills only installed for Claude Code. Now one command installs them for Codex and Cursor too, and it finds which tools you have. Some skills need Claude Code, so those skills now say so at the top. The setup steps also stopped writing files into another tool's folder.

---


richkuo:
@claude opus review

PR #141, opened 2026-08-14
feat(#138): add fable-dispatch skill with CLI shim for other harnesses [C45, Fable 5, high]

### LINKED ISSUE #138: [C45] Fable skills downgrade the model instead of using the Claude Code CLI on other harnesses
**Complexity: 45/100** — Capability 1 (Uncertainty 2 — dispatch shape needs discovery; Risk 2 — subprocess permission surface); Volume 20 — Opus 5, xhigh · fableplan: no

## Problem

The five skills that dispatch a Fable 5 subagent all assume the host harness has an Agent tool that accepts `model: fable`. When it does not, each one downgrades the model:

- `skills/fableplan/SKILL.md:124` — "fall back to the most capable model available"
- `skills/fable-validate/SKILL.md:73` — same wording
- `skills/fable-new-issue/SKILL.md:72` — same wording
- `skills/fable-advisor/SKILL.md:172` — same wording
- `skills/fable-orchestrate/SKILL.md:88` — "fall back to the closest available tier"

Two defects follow.

1. **The fallback is reactive.** All five sites condition on "(the Agent call errors on the model id)". A harness that accepts an unknown model id, or that maps `fable` to something else, produces a non-Fable plan with no error to catch.
2. **The fallback abandons Fable entirely.** `skills/fableplan/SKILL.md:106` already recognizes Cursor and Codex as supported harnesses for the worktree step, so these skills are expected to run there — and there they lose the one model the skill exists to use. A machine running Cursor or Codex may still have the Claude Code CLI installed; the ladder below tests for it rather than assuming it.

`skills/fableplan/SKILL.md:42` carries a related degradation: the skill checks whether the harness's Agent tool exposes an `effort` parameter and drops the stamped Plan effort when it does not. That path also has a CLI answer today.

## Goal

Every fable skill detects the harness before dispatching. On Claude Code it uses the Agent tool as it does now. On any other harness it shells out to the Claude Code CLI and still runs the work on Fable 5 at the intended effort. Downgrading to another model becomes the last resort, not the first, and the attribution footer names whatever actually ran.

## Approach

**Detection — positive, not reactive.** Test for Claude Code's own environment marker with a single Bash check: `[ -n "$CLAUDECODE" ]`. Verified present in a Claude Code session on this machine (`CLAUDECODE=1`, alongside `CLAUDE_CODE_ENTRYPOINT=cli`). Treat every other result as "not Claude Code". Do not test for Cursor or Codex variable names — those names are unverified here, and a positive test for the one harness we can confirm needs no knowledge of the others.

**CLI shim.** When detection says "not Claude Code", check `command -v claude`, then dispatch:

```
claude -p "<the same subagent prompt>" \
  --model fable --effort <tier> \
  --output-format json --permission-mode plan
```

All four flags are verified against the installed CLI: `--model` accepts the `fable` alias, `--effort` accepts `low|medium|high|xhigh|max`, `-p/--print` is non-interactive, and `--output-format json` returns an object whose keys include `result`, `is_error`, and `modelUsage`. A live `-p ... --output-format json --permission-mode plan` call was run against a cheap model and returned that shape.

- Read `.result` for the subagent's output and `.is_error` for failure.
- Read `.modelUsage` to confirm which model actually served the call, and name that model in the footer. This is what closes defect 1 — the shim observes the model instead of assuming it.
- `--permission-mode plan` keeps the Fable subagent read-only, matching what these skills already promise. `--dangerously-skip-permissions` must never appear in the shim.
- Pass the prompt through a file or a properly quoted argument. `tests/prompt-shell-safety.test.js` exists because a shell-interpolated prompt containing `"`, `` ` ``, or `$` breaks or executes; the shim must not reintroduce that class of bug.
- **Timeout.** A Fable 5 plan or validation at `high` effort runs for minutes, which can exceed a host harness's default Bash timeout and kill the call mid-run. The shared skill must require either the host's maximum Bash timeout or a background-and-poll pattern, and must state that the default timeout is not sufficient.

**Placement.** Add a new reference skill `skills/fable-dispatch/SKILL.md` holding the detection ladder, the shim command, the result-parsing contract, the failure and timeout rules, and the attribution rule. No existing skill is renamed or moved. Both installers discover skill directories automatically — `install.sh:24-26` symlinks every `skills/*/`, and `bin/install.mjs:20` picks up any directory containing a `SKILL.md` — so no manifest edit is needed. The five dispatching skills each replace their fallback paragraph with a one-line instruction to load `fable-dispatch` before dispatching. This matches the repo's existing shared-procedure pattern (`github-issue-format`, `pr-review-format`), which is the only cross-skill sharing mechanism that survives installation.

`workflows/milestone-pipeline.js` also names the `fable` model, but it runs only under Claude Code's own Workflow tool, so it is out of scope.

**Fallback order** (the shared skill owns it):

1. Claude Code harness → Agent tool with `model: fable`.
2. Not Claude Code, `claude` on PATH → the CLI shim above.
3. Neither, **or a shim call that fails** → fall back to the most capable available model, exactly as the current text says, and report the downgrade.

A shim call counts as failed when the CLI exits non-zero, when `.is_error` is true, or when `.modelUsage` shows a model other than Fable served the call. Report both the failure and the downgrade — a silent slide to step 3 is the same defect this issue fixes.

**Effort.** On the CLI path, `--effort` carries the stamped Plan effort directly, so `fableplan`'s "harness may not accept effort" degradation does not apply there. Keep the existing clamp: Fable never runs at `xhigh`, so a stamped `xhigh` becomes `high` and the skill says so.

Correctness of attribution outranks diff size here. Three call sites would be a smaller change than a new shared skill plus five edits, but five copies of a dispatch ladder drift, and a drifted copy produces a footer that names a model that did not run.

## Acceptance criteria

- [ ] `skills/fable-dispatch/SKILL.md` exists and defines: the `$CLAUDECODE` detection test, the three-step fallback ladder, the exact CLI command with `--model fable --effort <tier> --output-format json --permission-mode plan`, the `.result` / `.is_error` / `.modelUsage` parsing contract, and the rule that the footer names the model `modelUsage` reports.
- [ ] The shared skill defines shim-failure disposition: non-zero exit, `.is_error` true, or a non-Fable `.modelUsage` falls to ladder step 3 and reports both the failure and the downgrade.
- [ ] The shared skill defines a timeout rule for the shim — maximum Bash timeout or background-and-poll — and states that the default timeout is not sufficient for a Fable run at `high`.
- [ ] `fableplan`, `fable-validate`, `fable-new-issue`, `fable-advisor`, and `fable-orchestrate` each load `fable-dispatch` before dispatching, and none of them still carries its own inline "fall back to the most capable model" paragraph as the first response to an unavailable model.
- [ ] No existing skill directory, skill name, or trigger phrase changes.
- [ ] The shared skill forbids `--dangerously-skip-permissions` in the shim and requires a read-only permission mode.
- [ ] The shim passes the prompt without shell interpolation of `"`, `` ` ``, or `$`.
- [ ] A contract test (`tests/fable-dispatch-contract.test.js`, following the existing `tests/*-contract.test.js` pattern) asserts that every skill naming `model: fable` for a subagent dispatch references `fable-dispatch`, and that the shared skill's command line contains the four required flags and no dangerous-permission flag.
- [ ] `README.md`'s skill table gains a `fable-dispatch` row in the same reference-skill row style the table already uses for its other reference skills. Do not anchor on a line number — open issue #136 renames one of those rows.
- [ ] `bun test` passes.

---
Validated with LLM: Fable 5 | high | Harness: Claude Code | fable-validate

richkuo:
Closes #138

## Summary

- New reference skill `skills/fable-dispatch/SKILL.md`. It owns how every fable skill reaches Fable 5: positive harness detection (`[ -n "$CLAUDECODE" ]`), a Claude Code CLI shim for other harnesses, a three-step fallback ladder, the `.result` / `.is_error` / `.modelUsage` parsing contract, shim-failure disposition (non-zero exit, `.is_error` true, or a non-Fable `.modelUsage` → last-resort fallback with both the failure and the downgrade reported), a timeout rule (maximum Bash timeout or background-and-poll; the default timeout is stated as insufficient), and the attribution rule that the footer names the model `.modelUsage` reports.
- The shim command is `claude -p --model fable --effort <tier> --output-format json --permission-mode plan`, with the prompt fed via stdin from a file so `"`, backtick, and `$` are never shell-interpolated. `--dangerously-skip-permissions` is forbidden; a read-only permission mode is required. A stamped `xhigh` clamps to `high` on the CLI path with a notice.
- `fableplan`, `fable-validate`, `fable-new-issue`, `fable-advisor`, and `fable-orchestrate` each load `fable-dispatch` at their dispatch step. Their inline "fall back to the most capable model" paragraphs are replaced with pointers to the ladder. `fable-orchestrate`'s Sonnet-worker fallback stays local because worker dispatch is out of this issue's scope. `fableplan`'s effort-degradation note now states the degradation is Agent-tool-only.
- `tests/fable-dispatch-contract.test.js` pins the contract: every skill naming `model: fable` references `fable-dispatch`; the shim block carries the four required flags and no dangerous flag; detection, parsing keys, failure disposition, timeout rule, no-interpolation rule, and attribution markers are present; the five dispatchers carry no inline downgrade; README lists the row.
- `README.md` gains a `fable-dispatch` row in the existing reference-skill row style.
- No skill directory, skill name, or trigger phrase changed.

## Verification

- `bun test`: 180 pass, 0 fail (11 files). The new contract test failed before `skills/fable-dispatch/SKILL.md` existed and passed after (red → green).
- Live CLI check on this machine: `command -v claude` resolves; `--model`, `--effort`, `--output-format`, `--permission-mode`, and `-p` all appear in `claude --help`.
- Live shim call with a prompt file containing `"`, a backtick, and `$HOME`, fed via stdin: returned one JSON object with `is_error: false`, `result`, and `modelUsage`, and the special characters arrived as literal data. Notably the call requested `--model haiku` and `modelUsage` reported `claude-sonnet-5` served it — direct evidence for the issue's defect 1 and for the rule that the shim must read `.modelUsage` instead of assuming the requested model ran.

## Plain simple English

Some helper skills must run on the Fable 5 model. Before this change, they used a different model when the tool that runs them did not know Fable 5. Now a shared instruction file tells them to start Fable 5 through the Claude Code command line instead. A different model is only the last option, and it is always reported.

---


richkuo:
@claude fable review effort:medium

richkuo:
Addressed review feedback (Fable review via github-actions[bot] · 2026-08-14T19:53:32Z) in 9a92b5e.

### Fixed
1. **Dispatcher footers hardcode `Harness: Claude Code`** — all five dispatcher footer templates now use `<harness>`, filled with the harness actually running; `fable-dispatch` section 6 now covers the harness field beside model and effort (`skills/fable-dispatch/SKILL.md:59`). The same defect existed in `skills/fable-orchestrate/SKILL.md:79`, beyond the four cited files, and is fixed there too. A new contract test rejects a hardcoded harness in any dispatcher.
2. **Only `xhigh` clamps; `max` does not** (Recommended Optional 1) — the clamp now covers every tier above `high`: `xhigh`, `max`, and future tiers (`skills/fable-dispatch/SKILL.md:40`), test-pinned.

### Corrected scope (partial)
1. **Plan mode blocks the read-only `gh` calls in headless runs** — the categorical deny did not reproduce: a live headless `claude -p --permission-mode plan` run with no `--allowedTools` executed `gh pr list` with `permission_denials: []` and returned real data, so the permission classifier does approve read-only network commands here. The remedy still holds because classifier approval is heuristic and machine-dependent while an explicit allowlist is deterministic. The shim now carries a calling-skill-supplied `--allowedTools` read-only list, with the read-only-only constraint stated (`skills/fable-dispatch/SKILL.md:33,39`); a second live run with `--allowedTools "Bash(gh pr list *)"` confirmed the flag works in plan mode.

### Resolved judgment calls (was Recommended Optional 2)
1. **Ladder step 3 mechanism and the discarded substituted result** — implemented the suggested order: (a) the harness's own subagent facility, (b) adopt a completed shim result another model served, (c) inline with no writes so the calling skill's read-only promise holds (`skills/fable-dispatch/SKILL.md:24`). Substitution is now its own case in section 5: a completed call whose `.modelUsage` names another model is adopted as the step-3 result with the downgrade reported, because a re-run would land on the same substitute at extra cost (`skills/fable-dispatch/SKILL.md:55`). Alternatives rejected: always re-run on substitution (pays twice for the same downgrade); leave step 3 mechanism-free (unexecutable on harnesses without a subagent facility). Override if you prefer either.

Residual noted, out of this PR's scope: `fable-validate-fableplan`, `fable-validate-loop`, `fable-validate-fableplan-loop`, and the `fix-pr-review` templates also hardcode `Harness: Claude Code`; those files are untouched by this PR and unfiled.

Verification: `bun test` — 181 pass, 0 fail.

---

richkuo:
@claude review

richkuo:
Addressed review feedback (Opus 4.8 review via github-actions[bot] · 2026-08-14T20:11:24Z) in 33283bd.

### Fixed
1. **The read-only guarantee rested on a false plan-mode claim** — confirmed live, then corrected: a headless `claude -p --permission-mode plan --allowedTools "Bash(touch *)"` run executed the allow-listed `touch` and created its file, so plan mode does not block an allow-listed Bash write. The clause now states exactly that, and names the read-only-only constraint on the list as the shim's only guarantee against writes (`skills/fable-dispatch/SKILL.md:39`). The contract test rejects the old claim and requires the corrected one.
2. **Three fable-validate loop skills still hardcoded `Harness: Claude Code`** (Recommended Optional 2) — the same `<harness>` substitution applied to `skills/fable-validate-loop/SKILL.md:43`, `skills/fable-validate-fableplan/SKILL.md:47`, and `skills/fable-validate-fableplan-loop/SKILL.md:45`, each pointing at `fable-dispatch` section 6. The contract test now scans every fable-family skill for a hardcoded harness; the `fix-pr-review` templates stay untouched per the review.

### Resolved judgment calls (was Recommended Optional 1)
1. **No dispatcher names its `--allowedTools` read set** — implemented the review's second offered option: `fable-dispatch` now instructs the agent to derive the complete list from the calling skill's procedure, with `fable-validate`'s read set (`gh issue view --comments`, `gh pr list`) as the worked example, plus `git log`/`git diff` and `-R owner/repo` forms and the empty-list case (`skills/fable-dispatch/SKILL.md:39`). Alternative rejected: concrete per-skill lists in all five dispatchers — the procedures already name their reads, so duplicate lists would drift from the procedures they mirror and add five maintenance points for the same information. Override if you prefer explicit per-skill lists.

Verification: `bun test` — 181 pass, 0 fail.

---

richkuo:
@claude review

PR #146, opened 2026-08-14
feat(install): link skills and AGENTS.md into ~/.codex [C37, Opus 5, high]

richkuo:
## Summary

`install.sh` linked only into `~/.claude`. Codex read this repo's skills through hand-made symlinks in `~/.codex/skills`, so every skill added after those links were made stayed invisible to Codex, and the `pr-review-format` link kept dangling there after the rename.

- Extract the skill-link loop and the retired-name cleanup into `link_skills` and `retire_renamed_skills`, then run both against `~/.codex/skills` as well.
- Link `AGENTS.md` to `~/.codex/AGENTS.md` — the byte-identical twin of `CLAUDE.md` this repo already keeps in sync, and the file Codex reads for global instructions.
- Guard the whole Codex step on `~/.codex` already existing, so the script never plants a config directory for a tool the machine does not run.
- Keep workflows and the `/commit` command Claude-only; both are Claude Code formats with no Codex reader.

Backup behavior is unchanged: a real file or directory at a target moves to `<name>.bak`, and an existing `.bak` is never overwritten. The only outright deletions are symlinks.

## Verification

`bun test` — 207 pass, 0 fail, 14 files.

New `tests/install-codex-links.test.js` runs the real `install.sh` under a temporary `HOME` and asserts:

1. every shipped skill lands in `~/.codex/skills` as a symlink into the repo;
2. `~/.codex/AGENTS.md` links to the repo's `AGENTS.md`;
3. a real `~/.codex/AGENTS.md` moves to `.bak` instead of being discarded;
4. a dangling `pr-review-format` link in `~/.codex/skills` is retired;
5. nothing is created under `~/.codex` when Codex is absent, while `~/.claude` still installs in full;
6. no `workflows/` or `commands/` directory appears under `~/.codex`;
7. a backup an earlier run wrote survives a later run — a taken `.bak` falls through to `.bak.2` rather than being replaced or nested into;
8. with every backup name up to `.bak.99` taken, the file stays in place, unlinked, instead of clobbering one.

## Out of scope

`bin/install.mjs` (the `bunx rk-skills` path) is untouched and still installs to Claude destinations only.

## Plain simple English

The install script set up skills for Claude Code only. Codex used old links made by hand, so it did not see any newly added skill. The script now also sets up the same skills and the shared instructions file for Codex, but only when Codex is present. Existing files are kept as backups.

---

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-# Recreate the ~/.claude symlinks that point into this repo.
+# Recreate the ~/.claude (and ~/.codex) symlinks that point into this repo.
 # Safe: backs up any existing real file or directory to <name>.bak. The only
 # outright deletions are symlinks for the skills and agents this repo itself
 # retired, listed by name below — a symlink holds no data of its own.
@@ -7,6 +7,7 @@ set -euo pipefail
 
 REPO="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 CLAUDE="${HOME}/.claude"
+CODEX="${HOME}/.codex"
 
 link() {
   # link <repo-relative-source> <absolute-target>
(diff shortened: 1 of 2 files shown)


richkuo:
@claude fable review effort:medium

richkuo:
Addressed review feedback (@claude, Fable 5 · 2026-08-14T21:17:22Z) in 2012b25.

### Fixed

1. **`link()` overwrites an existing `.bak` backup** — confirmed against current code: `install.sh:19-21` ran `mv "$target" "$target.bak"` with no check, while `retire_renamed_skills` guarded the same case, so the file contradicted its own header comment. `link()` now resolves the first free backup name through a new `free_backup_path` helper — `.bak`, then `.bak.2`, `.bak.3` — so a backup an earlier run wrote is never replaced and never nested inside (`install.sh:20-38`, `install.sh:47-57`).

Two departures from the recommended fix, both toward the same invariant:

- **Next free name instead of keep-and-skip.** `retire_renamed_skills` keeps the directory in place when `.bak` is taken, and that is right for a *removal*: not retiring loses nothing. `link()` performs an *installation*, so the same behavior would silently leave the skill unlinked and the stale copy live. Falling through to `.bak.2` preserves every file and still completes the install. The retired-skill path keeps its guard exactly as it was — "must survive" item 3 — so neither path can clobber a backup.
- **Keep-and-skip survives as the exhaustion fallback.** After `.bak.99` the file stays in place, unlinked, with a printed reason. Refusing to link loses nothing; clobbering a backup loses a file.

Two details the trace turned up that the finding did not name:

- `free_backup_path` returns an explicit `0`. A bare `return` would carry the failed loop test's status out through the caller's `$(...)` assignment and trip `set -e`, aborting the install.
- The free-name test counts a dangling symlink as taken (`[ -e "$p" ] || [ -L "$p" ]`). `-e` alone reports a dangling link as absent, which would have written straight over it.

### Verification

`bun test` — 207 pass, 0 fail, 14 files. Four new cases in `tests/install-codex-links.test.js` cover the reviewer's "must survive" list: two successive runs each finding a fresh real file keep both backups (`.bak` and `.bak.2`); a directory target with a directory `.bak` present produces `.bak.2` with no nesting; exhausting all 99 names leaves the file in place, unlinked; and the pre-existing guard tests in `tests/pr-review-skill-name.test.js` still pass unchanged.

The PR body claim "an existing `.bak` is never overwritten" was true only of the retire path when the review ran. It now holds for both paths.

**Verification limitation noted in the review:** the Codex read paths (`~/.codex/AGENTS.md`, `~/.codex/skills`) could not be checked against OpenAI's documentation on that run — WebFetch was denied. They were confirmed empirically instead: this machine's `~/.codex/skills` already held working hand-made symlinks into this repo, which is what made the missing-skills problem visible.

---

richkuo:
@claude review

PR #165, opened 2026-08-17
docs(skills): cite the band-mapping and fable-effort owners instead of restating them [C2, Fable 5, high]

richkuo:
## Summary
- The complexity-band model/effort mapping now cites its owner, the validate-issue step 6 band table. milestoneplan and execution-plan-review each keep one inline copy that their table-rendering procedures need, marked with the owner as the authoritative source; every other repeat is a citation.
- The Fable effort ceiling (never xhigh) drops to one statement per file, citing the CLAUDE.md LLM Attribution Footer section as owner. prd-to-issues had three copies.
- milestone-workflow's Codex secret list now cites the fix-pr-review-loop review-bot preflight and keeps only the milestone-specific deltas; its duplicated cross-bucket failure row compresses to a pointer at step 1.

## Verification
- Re-read each touched file end to end. The surviving inline mappings match validate-issue step 6 exactly, and every unique rule survives.
- The task text placed the Codex secret list in fix-pr-review step 10; it actually lives in the fix-pr-review-loop preflight, so the citation names that file.

## Plain simple English

Several planning files repeated the same model-selection table and one rule about a model's effort limit. Now they point to the single files that own those rules. The rules did not change. There is one place to update them.

---

https://claude.ai/code/session_0165aPkQWpoao1cb5vGSVaek


richkuo:
@claude sonnet review

PR #173, opened 2026-08-21
feat(skills): require numbered plan steps with verify points [C8, Fable 5, medium]

richkuo:
## Summary

Big plans drift during long builds. This PR adds one contract between the plan producers and the plan consumer:

- `fableplan`, `issueplan`, `fable-advisor`, and `workflows/milestone-pipeline.js` (`planPrompt`) now instruct the planner to number its implementation steps and end each step with a verify point — the observable check (command, test, file state) that proves the step is done.
- `work-on-issue` step 2 now mirrors an adopted plan's numbered steps into the session task tracker before any code is written, for a plan of any length, and completes an item only when its verify point passes. Both fallbacks are specified: a plan lacking numbers or verify points gets both derived, and a harness with no task tracker carries the checklist in a scratchpad file outside the working tree.
- A step cancelled by one of step 2's three overrides closes as a **recorded deviation** carrying its own verify point, with a matching PR-body entry — so no mirrored item is left with an unreachable terminal state, and no tracker rule pressures a builder into building an overridden step.
- `fableplan` step 8, `issueplan` step 8, and `fable-advisor`'s prose path point at `work-on-issue` step 2, which owns the rule. `fable-advisor`'s issue path already inherits it through work-on-issue steps 1–6.
- `docs/contract-inventory.md` records the rule, its owner, its four producers and three pointer sites, guarded by `tests/plan-steps-contract.test.js`.

No splitting into sub-issues; the plan stays one artifact on the issue.

## Verification

- `bun test`: 242 pass, 0 fail (7 new assertions in `tests/plan-steps-contract.test.js`).
- Swept every plan-prompt site across the repo (`grep` for the shared plan-prompt phrasing and the `## Implementation plan` heading, over `*.md` and `*.js`, including `templates/`): four producers, all covered. An earlier revision of this body claimed three; the workflow prompt was the fourth.

## Plain simple English

When a plan for an issue is very long, the builder can lose its place and skip steps. Plans now come as a numbered checklist, and each step says how to prove it is done. The builder copies that checklist into its task list and marks steps done one by one. If a step turns out to be wrong, the builder closes it as a written change instead of leaving it open.

---
Updated with LLM: Opus 5 | high | Harness: fix-pr-review


richkuo:
@claude opus review

richkuo:
Addressed review feedback (@claude via github-actions[bot] · 2026-08-21T23:29:53Z) in `1a0e825`. All six findings re-validated against the code; all six confirmed.

### Fixed

1. **Overridden plan step has no disposition, and the mirror paragraph rebinds "This"** (Needs Fixing 1) — the mirror paragraph moved below the plan-deviation policy, so "This is the single plan-deviation policy" binds to the overrides again (`skills/work-on-issue/SKILL.md:93`). A new paragraph gives an override-cancelled step a reachable terminal state: it closes as a recorded deviation carrying its own verify point, with a matching PR-body entry per step 6; a step a newer issue comment deleted closes as removed with that comment as its record; two overrides on one step produce one disposition and one entry; and no tracker rule may push a builder into building an overridden step (`skills/work-on-issue/SKILL.md:102`).
2. **Fourth plan producer in the workflow script** (Optional 1) — `planPrompt` now asks for numbered steps with verify points and requires them in both the posted comment and the returned plan text; `PLAN_SCHEMA.plan` records that the numbering must survive; `implementPrompt` states the mirror rule alongside the plan handoff (`workflows/milestone-pipeline.js:282`, `:404`, `:437`). Confirmed the sweep myself over `*.md`/`*.js` including `templates/`: four producer sites, no fifth. The PR body's "three producers" claim is corrected.
3. **Three build paths produce numbered steps then never mirror them** (Optional 2) — `skills/fableplan/SKILL.md:98`, `skills/issueplan/SKILL.md:123`, and `fable-advisor`'s prose path (`skills/fable-advisor/SKILL.md:111`) now point at `work-on-issue` step 2 before writing any code. `fableplan`'s planning-phase-only mode still forbids steps 7–8, so it gains no build instruction; `fable-advisor`'s issue path keeps inheriting the rule through work-on-issue steps 1–6 and does not restate its substance. A guard asserts both.
4. **Both mirror fallbacks unspecified** (Optional 3) — a plan with no numbers or verify points now derives **both**, with an observable check per step, and a step whose check cannot run until a later step lands adopts that later check and stays open until then. A harness with no `TodoWrite`/`TaskCreate` writes the same numbered checklist to a scratchpad file, kept outside the repository working tree so step 5's staging cannot sweep it into the commit (`skills/work-on-issue/SKILL.md:95-98`).
5. **Guardrail row narrower than step 2** (Optional 4) — the row now triggers on "Adopted plan of any length, long or short", names the no-tracker fallback and the overridden-step disposition, and drops "long or many-part" (`skills/work-on-issue/SKILL.md:176`). A guard asserts the narrowing cannot come back.
6. **No contract-inventory row and no CI guard** (Optional 5) — added the "Numbered plan steps with verify points" row beside the adopted-plan family, with owner, the four producers, the three pointer sites, required shared semantics, allowed divergence, and its guard; added `fableplan`/`issueplan`/`fable-advisor` to the inventoried-files table; scoped the out-of-scope table to the loop/validate guard so `planPrompt` stays an in-scope consumer of this row (`docs/contract-inventory.md:3`, `:11`, `:34-37`, `:55`, `:74`). New guard `tests/plan-steps-contract.test.js` asserts shared semantics, not exact strings. The divergence you flagged is closed: `skills/fable-advisor/SKILL.md:65` now carries the verify-point examples and the builder-mirror reason.

### Verification

`bun test`: 242 pass, 0 fail (was 235). `node --check` on `workflows/milestone-pipeline.js` reports a top-level `return` — that is pre-existing on `main` and by design for the workflow runtime, confirmed by running the same check against the unmodified base.

---

richkuo:
Correction to three `file:line` anchors in the disposition above — the substance is unchanged, only the line numbers were off:

- Item 1: the plan-deviation policy sits at `skills/work-on-issue/SKILL.md:93`; the mirror paragraph that moved below it is at `:95`.
- Item 6: the new inventory rule row is `docs/contract-inventory.md:54` (I wrote `:55`, which is the adopted-plan deviation row it sits above). The three added inventoried-files rows are `:32-34` (I wrote `:34-37`). The out-of-scope scoping note is `:75` (I wrote `:74`, a blank line).

---

richkuo:
@claude review

richkuo:
Addressed review feedback (@claude via github-actions[bot] · 2026-08-21T23:46:51Z) in `d329618`. One finding, re-validated against the code and confirmed.

### Fixed

1. **A step that borrowed a later step's verify point had no reachable terminal state when an override removed that later step** (Recommended Optional 1) — traced it and confirmed the gap: `skills/work-on-issue/SKILL.md:97` lets an item adopt a later step's check, `:102` disposes only the overridden step itself, so a borrowing item stayed open with a check that can never run, which `:100` defines as the signal for a skipped step. A new paragraph at `:104` gives the borrower a terminal state: when an override cancels, replaces, or removes the step whose check was borrowed, **every** open item that borrowed it re-homes its check — to the replacement step's verify point when the override replaced the step, otherwise to its own observable check, and only when no such check can exist does the item close as a recorded deviation with its own PR-body entry per step 6. Your three must-survive cases hold: a deleted step 7 leaves item 3 with its own check or a deviation; a replaced step 5 hands item 3 the replacement's check; two borrowers both re-home, because the rule says "every". The first fallback bullet at `:97` now points at that disposition instead of restating it, so the file keeps one statement of the rule.

`tests/plan-steps-contract.test.js` gains a guard asserting the four shared semantics (every borrower re-homes, a replacement's check is followed, a borrower without a replacement still terminates, no borrower waits on an unreachable check) as patterns, not exact strings. The `docs/contract-inventory.md:54` row records the re-home clause in its required-shared-semantics column.

### Verification

`bun test`: 243 pass, 0 fail (was 242). The new guard was proven red → green — with the new paragraph removed the guard fails, and it passes with the paragraph restored.

---

richkuo:
@claude sonnet review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-08-21T23:58:00Z) in 18bd360.

### Fixed

1. **Re-homing protected only agent-derived borrowed checks, not author-written ones** — confirmed at `skills/work-on-issue/SKILL.md:104`: the trigger read "adopted a later step's check **under the first fallback above**", so a compliant numbered plan whose own verify point says "proven once step 7's test exercises it" never reached the rule, because the no-numbers fallback never fires for it. The paragraph now triggers on the cross-reference itself — "derived by the mirroring agent under the first fallback above, or written by the plan's own author" — and names the copy-verbatim path that produces it.

2. **Chained borrows did not resolve transitively** — the `Must survive` case where step 3 borrows step 5's check and step 5 borrows step 8's: overriding step 8 re-homed step 5, but if step 5 then closed as a deviation of its own, step 3 was left keying on a check that can never run. `skills/work-on-issue/SKILL.md:104` adds **Re-homing cascades**: when a borrowing item closes as a deviation, every item that borrowed *that* item's check re-homes in turn, and the chain repeats until no open item keys on an unreachable check.

3. **The guardrail table gave no borrowing signal** — confirmed: the table at `skills/work-on-issue/SKILL.md:176` mentioned only the overridden-step disposition, so an agent reading the table alone could close a borrowing item as a bare deviation the moment its source step was overridden. A new row states the re-homing order, its origin-agnostic scope, and the cascade, and forbids both closing the item as done and leaving it on an unreachable check.

4. **Inventory and owner text disagreed on scope** — `docs/contract-inventory.md:54` stated the guarantee unconditionally while the owner scoped it to the derived fallback. The inventory's required shared semantics now carries the same origin-agnostic wording and the cascade clause, so the two agree.

### Verification

`tests/plan-steps-contract.test.js` gains two guards: origin-agnostic scope plus the cascade inside the existing borrowing test, and a new test for the guardrail row. Both were run against the pre-fix text first and failed there (`re-homing is origin-agnostic`, `borrowing guardrail row present`) — red before green. Full suite after the fix: 244 pass, 0 fail.

## Plain simple English

A plan step can say it is proven when a later step's test passes. The rule that protects that link applied only when the checklist invented the link. Now it applies also when the plan's author wrote it, and it follows chains of such links. The guardrail table and the contract inventory say the same thing.

---

richkuo:
@claude sonnet review

richkuo:
Addressed review feedback (@claude via github-actions[bot] · 2026-08-22T00:09:41Z) in `17ffa63`. Three findings, each re-validated against the code and confirmed.

### Fixed

1. **The mirror fallback's trigger did not clearly cover a plan missing only one element** (Needs Fixing 1) — confirmed at `skills/work-on-issue/SKILL.md:97`: the trigger read "The plan carries no numbers or verify points ... derive **both**", which parses as AND, so a numbered plan with no per-step checks ("1. Add endpoint. 2. Add tests.") never reached the fallback while the same bullet's closing clause forbids leaving an item with no verify point. The bullet now triggers on "missing its numbers, its verify points, or both", names the numbered-plan-without-checks case in its examples, derives whichever element is absent, and states explicitly that either element missing on its own fires the fallback. A plan missing both still gets both. `docs/contract-inventory.md:54` carries the same wording, so the recorded contract no longer restates the ambiguity.

2. **The fallback's forward pointer named the wrong paragraph** (Recommended Optional 1) — confirmed: "the overridden-step disposition below owns that case" matched the title of `:102`, which disposes the cancelled step's *own* item and carries no replacement-check-first ordering, while `:104` owns the borrowed-check case. The parenthetical now names "A borrowed verify point re-homes when its source step is overridden" by title, spells out its ordering (replacement's check, else the item's own check, else a deviation of its own), and says plainly that `:102` governs the cancelled step's own item rather than its borrowers. A reader following the literal pointer now lands on the ordered rule, so the replaced-step case resolves correctly.

3. **The workflow plan producer omitted the plan-comment heading** (Recommended Optional 2) — confirmed: `planPrompt` was the only one of the four producers with no heading instruction, while `work-on-issue` step 0 (`skills/work-on-issue/SKILL.md:41`) matches on `## Implementation plan` first and falls back to a weaker "a maintainer clearly frames it" bar that an automated bot comment does not reliably meet. `workflows/milestone-pipeline.js:406` now requires the `## Implementation plan (Fable 5)` heading above the plan body and states why, so a later standalone `work-on-issue` run on a milestone-planned issue finds the plan. The milestone pipeline's own implement agent is unaffected, because it is told directly that a plan exists.

### Verification

`tests/plan-steps-contract.test.js` gains three guards, asserting shared semantics as patterns rather than exact strings: the fallback trigger covers either element and the superseded AND-only wording cannot return; the fallback pointer names the re-homing rule and distinguishes the two dispositions; every one of the four producers asks for the `## Implementation plan (` heading and the owner still matches on it. All three were run against the pre-fix text first and failed there — `fallback trigger covers either element`, `fallback points at the re-homing rule`, `plan-comment heading` — one failing guard per finding, red before green. Full suite after the fix: 245 pass, 0 fail (was 244). `docs/contract-inventory.md:54` records both new semantics per the sync rule at line 12.

## Plain simple English

A plan can have numbers but no checks, or checks but no numbers. The rule that adds the missing part only fired clearly when both parts were missing. It now fires when either part is missing. One sentence pointed at the wrong rule, so it now names the correct one. The workflow file now asks for the same plan heading as the other three places.

---

richkuo:
@claude review

PR #174, opened 2026-08-22
feat(skills): lead the validate verdict with the recommended next step [C12, Opus 5, high]

richkuo:
## Summary

The `validate-issue` step-8 verdict block ended with one static options line, so a verdict that answered "Update issue description? **Yes**" still gave no recommendation. This adds a **Next-step line** rule that leads with the recommended action, then lists the other options:

1. Scope is split or umbrella → recommend `split issue`
2. Update issue description is Yes → recommend `update issue`
3. Otherwise → the existing `Reply "work on issue" to proceed...` line

The `"fableplan" when offered` option drops out of the line when the fableplan signal is no.

Every Fable and loop wrapper (`fable-validate`, `fable-validate-fableplan`, `validate-issue-loop`, `fable-validate-loop`, `validate-fableplan-loop`) presents the verdict in the validate-issue step-8 format, so they inherit this rule with no change of their own.

## Verification

- `bun test` — 245 pass, 0 fail
- `grep` over the repo for the old literal line: the two remaining references (`work-on-issue`, `validate-issue-loop`) now name the next-step line generically instead of quoting the old text

## Plain simple English

When a validation says an issue needs edits, the last line now tells you to reply "update issue" first. Before, it always showed the same list of choices and gave no recommendation.

---

### DIFF
--- a/skills/validate-issue-loop/SKILL.md
+++ b/skills/validate-issue-loop/SKILL.md
@@ -5,7 +5,7 @@ description: Use when the user asks to validate a GitHub issue and then autonomo
 
 # validate-issue-loop
 
-Chain validate-issue → (conditional) update issue → work-on-issue-loop into one autonomous run, so an issue goes from "reported" to "PR through N rounds of review" without a human in the loop between steps. This is validate-issue's normal interactive handoff (`→ Reply "work on issue"`) made unattended: the loop reads its own verdict and decides what to do next, instead of waiting for the user to type a reply.
+Chain validate-issue → (conditional) update issue → work-on-issue-loop into one autonomous run, so an issue goes from "reported" to "PR through N rounds of review" without a human in the loop between steps. This is validate-issue's normal interactive handoff (its `→` next-step line) made unattended: the loop reads its own verdict and decides what to do next, instead of waiting for the user to type a reply.
 
 **Do not skip validation.** Auto-implementing an issue whose factual claims or proposal you haven't traced against the code just reproduces the issue's own mistakes in a PR. Every step of validate-issue still runs; only the "wait for the user's reply" step is replaced by a decision table.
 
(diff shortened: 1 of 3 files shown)


richkuo:
@claude sonnet review

richkuo:
Addressed review feedback (@claude · 2026-08-22T07:27:04Z) in 3e4affe.

### Fixed
1. **Next-step line template duplicates the arrow** — confirmed: the template placeholder at `skills/validate-issue/SKILL.md:127` carried its own `→` while all three substitution values also start with `→`. The placeholder is now `<next-step line — see the rule below; it carries its own leading arrow>`, so each case renders exactly one arrow (`skills/validate-issue/SKILL.md:127`).
2. **Dropping "fableplan" leaves a dangling conjunction** — confirmed: the phrase-removal rule produced `, or .` in case 3 and a trailing `, .` in cases 1–2. The rule is replaced by explicit fableplan yes/no variants for all three cases, each a complete sentence posted verbatim, plus a line forbidding assembly by word deletion (`skills/validate-issue/SKILL.md:132-144`).

### Resolved judgment calls
1. **Wording of the fableplan option** — while rewriting these lines, `"fableplan" when offered` became `"fableplan" to plan first`. With per-signal variants, the yes variant *is* the offered case, so "when offered" was redundant, and the new wording is parallel with the other options ("to restructure", "to edit", "to build as-is"). Alternative rejected: keep "when offered" — it reads as a condition the user must evaluate, in a line that is already conditional on the signal. Override if you prefer the original phrase.

Verification: `bun test` — 245 pass, 0 fail. All six variants read as complete sentences with a single leading arrow; no other skill parses this wording (`skills/validate-issue-loop/SKILL.md:26` explicitly treats the verdict block, not the arrow line, as its input).

---

richkuo:
@claude review

PR #177, opened 2026-08-22
feat(review-bands): add a sonnet first-review band and move the boundaries [C38, Opus 5, high]

richkuo:
## Summary

Re-scales the first-review band table from three tiers to four, adding a Sonnet tier below the standard trigger:

| Band | Trigger |
|---|---|
| C0–C10 | `@claude sonnet review` |
| C11–C40 | `@claude review` |
| C41–C80 | `@claude opus review` |
| C81+, or no score | `@claude fable review effort:high` |

The C81+ step-down ladder keeps its shape — Fable reviews cycle 1, then Opus, then the standard trigger — and now floors explicitly at `@claude review`. A C81+ pull request never steps down to Sonnet, however many cycles it takes. Sonnet is a band tier and the non-blocking tier; it is never a C81+ blocking rung.

On Codex, the C0–C10 band maps to the cheap shorthand (`@codex luna review`) alongside the non-blocking re-review. C11 and above still collapse onto the bare `@codex review`.

`validate-issue` step 6 remains the authoritative table.

## Verification

`bun test` — 249 pass, 0 fail.

Band assertions in `tests/complexity-score.test.js`, `tests/pr-review-contract.test.js`, and `tests/milestone-pipeline.test.js` were moved to the new boundaries and extended to cover the Sonnet band and the ladder floor. The milestone-pipeline fixtures now straddle 10/11, 40/41, and 80/81.

## Dependent sweep

Every restatement of the table was updated: `fix-pr-review`, `fix-pr-review-loop`, `work-on-issue-loop`, `milestone-workflow`, `milestoneplan`, `prd-to-issues`, `validate-issue` and its `complexity-scoring.md`, both Action prompt templates, `README.md`, `docs/contract-inventory.md`, and `REVIEW_BANDS` in `workflows/milestone-pipeline.js`. A repository-wide search finds no remaining reference to the old boundaries.

## Side effect: fix-pr-review step 10 split out

The added band row pushed `skills/fix-pr-review/SKILL.md` to 201 lines, one over the 200-line ceiling that `tests/pr-review-contract.test.js` enforces. The routing procedure is lookup data, so it moved to `skills/fix-pr-review/rereview-routing.md`, following the same reference-file pattern as `fetch-recipes.md` and `disposition-comment.md`.

`SKILL.md` is now 175 lines. Step 10 keeps a pointer plus the two rules that decide the routing: route by whether the addressed set contained a blocking finding, and post the trigger as its own comment. The size test was extended to require the new reference and to assert the routing file carries the band table, the ladder floor, and both bots' commands.

## Plain simple English

Pull request reviews now choose the reviewer from four complexity levels instead of three. Very small changes get a cheaper reviewer. Very large changes still get the strongest reviewer first, and the reviewer never falls below the standard one, however many review rounds it takes.

---


richkuo:
@claude opus review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-08-22T23:25:18Z) in e248887.

### Fixed

1. **Codex Action prompts post a bare `@codex review` for a C0–C10 PR** — confirmed; both prompts flattened every band onto the flagship with no band read. `templates/codex-workflow/prompts/issue-workflow.md:11` now routes the first review by the issue's `[C<score>]` prefix, and `templates/codex-workflow/prompts/fix-pr.md:19` routes the blocking re-review by the PR title bracket (falling back to the closed issue's prefix): `@codex luna review` at C0–C10, the bare `@codex review` at C11 and above. Both state that the C81+ ladder stays on the bare trigger and never reaches luna. Verified the shorthand is postable today — `templates/codex-workflow/workflows/codex.yml:162` resolves `luna` to `gpt-5.6-luna`. Also corrected `templates/codex-workflow/README.md:130`, which still described luna as only the cheap re-review shorthand, and named both Codex prompts in the consumer column of `docs/contract-inventory.md:46`.

2. **`complexity-scoring.md` contradicts its own skill's table on Sonnet** — confirmed. `skills/validate-issue/complexity-scoring.md:39` told the validator that Sonnet appears only for a cheaper non-blocking re-review, while `skills/validate-issue/SKILL.md:100` now assigns Sonnet 5 · high to the 0–10 first review and `workflows/milestone-pipeline.js:206` encodes the same. The boundary sweep missed the line because it carries no boundary numbers. Corrected it to name Sonnet as both the 0–10 first review and the non-blocking re-review in every band. Your second point on the adjacent bullet was also right: `complexity-scoring.md:38` was the last step-down restatement in the repo omitting the ladder floor, so it now states that the ladder stops at the standard reviewer and never reaches Sonnet.

### Corrected scope (partial)

1. **"Add both Codex prompts to the `consumers` list in `tests/pr-review-contract.test.js:324-330`"** — the gap is real and is now covered, but not by that list. That loop asserts `@claude opus review`, `@claude sonnet review` and `C0.{0,4}C10` against every member (`tests/pr-review-contract.test.js:340-342`), and the Codex prompts must never name a `@claude` trigger — the codex-mode assertion at `tests/milestone-pipeline.test.js:1073` requires "never switch to @claude". Adding them to that list would have failed on the first three assertions. Instead there is a sibling test, `the Codex Action prompts route the cheap band to @codex luna review` (`tests/pr-review-contract.test.js:366-390`), asserting the cheap band, the C11+ collapse, the absence of any `@claude` trigger, and the C81+ ladder floor. Verified it fails against the pre-fix prompt and passes after, so it genuinely closes the hole rather than only appearing to.

All three of your "must survive" cases hold: a C0–C10 Codex first review posts `@codex luna review`; a C0–C10 Codex blocking re-review with a `Needs Fixing` item posts `@codex luna review`; a C81+ Codex PR at cycle 3 or later stays on the bare `@codex review`.

Tests: 250 pass, 0 fail (`bun test`), up from 249 with the new case. CI `test` was already green on the reviewed head and no CI findings were raised.

## Plain simple English

The Codex robot prompts asked for the strong reviewer on the smallest changes. They now ask for the cheap reviewer there, and keep the strong one everywhere else. One helper file also gave the validator a rule that its own table contradicted, and that is corrected. A new test stops the same gap from coming back.

---

richkuo:
@claude opus review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-08-22T23:57:28Z) in 7650fec.

### Fixed

1. **Step 1 sends `Create Follow-up Issue` filing to step 6, which orders those items skipped** — confirmed, and the cause is exactly as traced. Commit `3a3a10b` compressed the paragraph and widened `(step 6)` from the `Recommended Optional` clause alone to cover both clauses; `skills/fix-pr-review/SKILL.md:112` skips `Create Follow-up Issue` items, and step 1's LGTM-only path is terminal, so the deferred items were lost with no record on the PR. `skills/fix-pr-review/SKILL.md:53` now reads "Implement each `Recommended Optional` item per step 6, and file each `Create Follow-up Issue` item as an issue per step 3's bar — step 6 edits no code for it." All three of your must-survive cases hold: a bare `LGTM` carrying only follow-ups files each one before reporting approved; a `Needs Updates` review's follow-up still files through the disposition's deferred section with no step 6 code edit; a mixed `LGTM` routes the optional item to step 6 and the follow-up to step 3.

2. **The cycle-1 implement prompt ordered a flat re-trigger, contradicting the cycle-2+ prompt in the same file** — confirmed, and your sequence trace is right: a `[C10]` Codex issue got `@codex luna review` as its first review (`firstReviewTrigger`, `CODEX_REVIEW_SHORTHAND` at `workflows/milestone-pipeline.js:228`), the bare flagship from the cycle-1 agent at line 449, then `@codex luna review` again from the cycle-2 agent at line 472. Both branches of line 449 now derive the blocking re-trigger from the band. The Claude branch carried the mirror gap you identified: it never named the C0–C10 sonnet tier, and its sonnet ban was unscoped, so a `[C8]` reader could take it as a blanket ban on the trigger its own band requires. It now names the tier and scopes the ban to "that C81+ ladder", matching line 472's phrasing.

3. **Recommended Optional — the routing reference kept the old two-tier Codex count and stated the Codex mapping only in a shell comment** — confirmed and implemented. Under this PR three Claude tiers collapse onto the bare `@codex review`, and `skills/fix-pr-review/rereview-routing.md:62` was the last survivor of the old phrasing. Section 2's band table now has a **Codex trigger** column (`@codex luna review` at C0–C10, bare `@codex review` at C11–C40, C41–C80 and at every C81+ cycle) plus a rule line stating that the ladder never reaches `luna`. A `/fix-pr-review <N> codex` run now gets its trigger from the normative prose rather than a shell comment. Your three must-survive cases read straight off the new column.

### Test coverage

You correctly noted `tests/milestone-pipeline.test.js:1070-1079` covered only the first-review trigger and the "never switch to @claude" string, leaving both branches of the step-4 parenthetical untested. The codex case now asserts the band-derived re-trigger and the "never reaching luna" clause, and a new case, `the github-mode cycle-1 prompt derives its Claude re-trigger from the band`, covers the Claude branch at `[C8]` and `[C90]`. Verified both fail against the pre-fix prompt and pass after.

Tests: 251 pass, 0 fail (`bun test`), up from 250. CI `test` was green on the reviewed head; no CI findings were raised.

**Verification limitation noted, not treated as a finding:** your run could not reach the OpenAI platform model list and confirmed the `luna` → `gpt-5.6-luna` mapping in-repo instead. I did not independently verify the external model list either, so that mapping rests on the same in-repo evidence.

## Plain simple English

One line in the skill told the agent to file follow-up issues in a step that skips them, so that work was lost. It now points at the step that files them. A pipeline prompt also asked for the strong reviewer on the smallest changes while a second prompt in the same file asked for the cheap one. Both now read the complexity band.

---

richkuo:
@claude opus review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-08-23T00:14:31Z) in 797218d.

### Fixed

1. **`fix-pr-review-loop` step 1 ordered a flat `@codex review` for every band** — confirmed. `skills/fix-pr-review-loop/SKILL.md:45` named a fixed trigger while line 43 made the mapping band-dependent, and `skills/work-on-issue-loop/SKILL.md:32` inherits its Codex trigger entirely from that line. Line 45 now takes the trigger from the mapping above it and states both bands explicitly. I also verified your `@codex sonnet review` claim against `.github/workflows/codex.yml:199` and `templates/codex-workflow/workflows/codex.yml:199`: the shorthand alternation is `sol|terra|luna|mini|codex|spark`, so that phrase matches no branch of the command regex, and the line now says so. Your three must-survive cases hold. New test `no site names a fixed Codex trigger that overrides the band` covers all four Codex trigger sites, and `no site maps a @claude model shorthand onto @codex` guards ten sites against composing one, with the shorthand set asserted against both workflow files so the guard cannot rot.

2. **Three restatements keyed the step-down ladder to the C81+ band while the code keys it to a Fable first review** — confirmed, and your trace is right. `MODEL_IDS` is an identity map (`workflows/milestone-pipeline.js:176`), so `firstReview.model` at `:596` really does hold `'fable'` for a stamp, and `:605` steps down at any score. All the band-keyed prose is now Fable-keyed: `skills/fix-pr-review/rereview-routing.md` (its blocking section now checks what cycle 1 ran **before** the band table, and the ladder heading no longer says C81+), `skills/fix-pr-review-loop/SKILL.md:41`, `skills/validate-issue/SKILL.md:105`, `skills/validate-issue/complexity-scoring.md:38`, `README.md:113`, `docs/contract-inventory.md:46`. I extended the sweep to `templates/claude-workflow/prompts/fix-pr.md:17`, which carried the same defect — it detected the fable trigger comment by event but gated that detection behind the C81+ band, so a stamped Fable at C55 fell through to the C41–C80 row. It now checks for the fable trigger comment first and falls back to the band only when none is present. New test `every step-down statement keys the ladder to a Fable first review, never to a band` asserts the band-keyed phrasings are gone from all seven sites and that each states the cycle-1 key.

### Resolved judgment calls

None — both blocking findings had a single correct remedy once traced.

### Recommended Optional — implemented

1. **The Codex first-review prompt dropped the stamped-override rule its Claude sibling keeps** — confirmed and implemented. `firstReviewTrigger` at `workflows/milestone-pipeline.js:238-244` does honour a stamp on the Codex route, so the prompt was the only site discarding it. `templates/codex-workflow/prompts/issue-workflow.md:11` now maps a stamped model onto the Codex column — `luna` for sonnet/haiku, the bare trigger for opus/fable — and carries the `effort:<tier>` suffix, which `codex.yml:123` accepts on the trigger gate and `:172` extracts. Your three must-survive cases read straight off the new clause. New test `the Codex first-review prompt honours a stamped PR review line`, which also asserts the Claude sibling still states the override so the pair stays in step.

2. **A genuine `[C0]` routed to Fable because the pipeline could not tell a real zero from a missing prefix** — confirmed, and `[C0]` is producible: `25 × 0 + 0 = 0` when all five axes grade 0. I took your cleanest fix rather than the documentation-only alternative, because the invariant you wrote is the correct one and the narrower option leaves the authoritative table disagreeing with its encoding. `complexity` is now optional in `PREP_SCHEMA` (`workflows/milestone-pipeline.js:271`), the prep prompt tells the agent to omit it for an unprefixed title (`:667`), and a single `hasScore` predicate at `:217` drives both `bandFor` and `reviewBandFor`, so build and review routing cannot disagree about what "unknown" means.

   One thing your write-up did not raise, which I would not ship without: relying on the agent's omission alone **inverts the failure direction**. Today a prep slip escalates, which is safe; under a bare optional field a slip would route an unknown-complexity issue to the *cheapest* band. So the runtime now reconciles the one ambiguous value against the title it already holds (`:681-691`): a reported `0` stands only when the title really reads `[C0]`, and otherwise drops to unknown with a log line. That reconciliation can only escalate, so a title outside the convention behaves exactly as before. I also had to guard the rescore fallback at `:859` — `rescored` defaulted to `0` for "could not score", which under the new semantics would have collapsed every unscored, unrescored issue onto the bottom band. It is now `undefined` in that case.

   New tests `a literal [C0] is a real score, and only an absent prefix routes as unknown` (your three cases plus the prep-slip guard) and `the prep contract tells a real zero from a missing prefix` (the schema, prompt and predicate encoding).

### Not changed

Nothing was refuted. All four items traced true.

### Test coverage

Six new cases. Every one was verified failing against the pre-fix tree (`git stash` of the source files, tests held) and passing after. Tests: 258 pass, 0 fail (`bun test`), up from 251. CI `test` was green on the reviewed head; no CI findings were raised.

**Verification limitations noted, not treated as findings:** your run could not reach the OpenAI platform model list or the Anthropic model overview, and confirmed the `luna` → `gpt-5.6-luna` and `sonnet` → `claude-sonnet-5` mappings in-repo instead. I did not reach either external source either, so both mappings rest on the same in-repo evidence.

## Plain simple English

Some pages said the reviewer steps down only for very large changes. The code steps down whenever the first review used Fable, which an operator can choose at any size. The pages now agree with the code. One line also told the agent to call the plain Codex reviewer for every size, which contradicted the line above it. Finally, an issue scored zero now gets the cheapest reviewer, as its table always said.

---

richkuo:
@claude opus review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-08-23T00:34:33Z) in 0e8b0c6.

### Fixed

1. **Four re-review sites routed a blocking re-review by band alone** — added `blockingRetrigger()` (`workflows/milestone-pipeline.js:266-273`), which reads the cycle-1 trigger `firstReviewTrigger(ex)` returns and steps down only when it names Fable, mirroring `runSubagentReviewLoop`'s `firstReview.model !== 'fable'` guard. Both pipeline prompt sites now emit the exact computed trigger instead of a band list: the cycle-1 directive names `the blocking re-trigger exactly <trigger>` (`workflows/milestone-pipeline.js:462`), and the batch prompt names the cycle-1 trigger and orders the agent to confirm it against the earliest `@<bot> … review` comment before relying on it (`workflows/milestone-pipeline.js:485`). Both Action prompts got the same procedure — `templates/claude-workflow/prompts/fix-pr.md:17` now identifies the cycle-1 reviewer from the earliest `@claude … review` comment, takes the ladder only when that names fable, repeats any other model verbatim, and reaches a band row only when the PR carries no trigger comment at all; `templates/codex-workflow/prompts/fix-pr.md:19` states the same identification step and repeats the cycle-1 trigger, since Codex has one flagship and no Fable tier. The C81+ Claude band row now reads `@claude opus review` rather than the bare trigger, because a Fable cycle never opens on a re-review. Consumer lists extended to `workflows/milestone-pipeline.js` and `templates/codex-workflow/prompts/fix-pr.md` (`tests/pr-review-contract.test.js:475-484`). All three "Must survive" cases are now pinned as assertions (`tests/milestone-pipeline.test.js:1192-1229`): a `[C5]` stamped `@claude opus review effort:high` keeps that trigger at every blocking cycle; a `[C90]` stamped opus carries no fable comment and keeps opus rather than the band's bare trigger; a `[C60]` Codex issue stamped sonnet opens on `@codex luna review effort:high` and stays there, while an unstamped `[C90]` Claude issue keeps the Fable → Opus ladder.

2. **The prep guard tested prefix presence, not the score value** — `workflows/milestone-pipeline.js:692-700` now captures the prefix with `SCORE_PREFIX.exec` and reconciles **any** reported score against it, not just a reported 0. Any disagreement drops to unknown, which takes the top band, so the guard can only escalate. The log line names the real prefix (`the title reads [C50]`) so the slip is diagnosable. Covered by three new fixtures (`tests/milestone-pipeline.test.js:388-396`): `[C50]` with a reported 0 and `[C90]` with a reported 5 both route as unknown on the top band, `[C0]` with a reported 0 stays a real zero and emits no log, and the unprefixed and `[Cx]` cases still produce no spurious reconciliation. The fixture titles across `tests/milestone-pipeline.test.js` were corrected to carry the score they report — a prep result pairing `complexity: 10` with a prefix-less title is itself the slip this guard catches, so the old fixtures were not reachable states.

3. **`fix-pr-review-loop` understated what `@codex sonnet review` does** — corrected at `skills/fix-pr-review-loop/SKILL.md:45`. The rule against carrying a `@claude` shorthand to `@codex` stays; the consequence now matches the Action. Traced: `.github/workflows/codex.yml:130` sets `invoked=true` for any line-start `@codex`, `:154` finds no `sonnet` arm so `:165` falls through to `gpt-5.6-sol`, and `:272-274` takes the first token that is not one of the six shorthands as the route keyword, giving `KEYWORD=sonnet`; `:285-291` then selects `MODE=fix-pr` on a trusted-author PR (OWNER/MEMBER/COLLABORATOR or `CODEX_BOT_LOGIN`), the route that mints the App token and pushes. The fork / untrusted-author carve-out is stated too, since that path still degrades to read-only review. The test that pinned the wrong wording was rewritten to assert the corrected claim and to forbid the old one (`tests/pr-review-contract.test.js:468-483`).

4. **An unscored issue rescored low took the cheapest first review** (Recommended Optional) — added `ex.review_complexity` (`workflows/milestone-pipeline.js:893`), clamped upward-only the way the validate band at `:872` is: an issue with no `[C<score>]` prefix keeps unknown for review routing, because nothing outranks the top band, while a scored issue keeps the escalated value. Both read sites now use it (`workflows/milestone-pipeline.js:247`, `:611`). An unprefixed issue whose validator returns 5 or 90 stays on the top band; a `[C5]` issue rescored to 90 still escalates to the 81+ review band.

Verification: `bun test` — 258 pass, 0 fail.

**Plain simple English:** An operator can name the reviewer for a pull request. Four files chose the next reviewer from the size band and forgot that choice, so a strong reviewer dropped to the cheapest one. Each file now reads the reviewer that ran the first round. A score check now compares the number, not only its presence.

---

richkuo:
@claude opus review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-08-23T00:55:43Z) in 4c89323.

### Fixed

1. **No skill site mapped a stamped `PR review:` model onto the Codex column** — confirmed at HEAD: `skills/fix-pr-review-loop/SKILL.md:43` stated the Codex column by band only, and `:45` ordered the trigger taken from that mapping with no stamp rule, while `:41` says a stamp overrides the band. `skills/fix-pr-review/rereview-routing.md` had the same hole: `:19` keyed the step-down to a stamp naming Fable, the ladder at `:38-42` carried `@claude` rungs that `:7` forbids on a Codex cycle, and the fallback Codex column at `:23-28` was band-only. Both pages now carry the mapping — a stamped `sonnet` or `haiku` becomes `@codex luna review`, a stamped `opus` or `fable` becomes the bare `@codex review`, each keeping the stamped `effort:<tier>` (`skills/fix-pr-review-loop/SKILL.md:43,45`; `skills/fix-pr-review/rereview-routing.md:21`). The ladder is now scoped to Claude cycles and states that Codex has no Fable tier, so its cycle-1 trigger repeats for every blocking re-review (`skills/fix-pr-review/rereview-routing.md:36-40`). `skills/work-on-issue-loop/SKILL.md:30,32` delegates to those lines rather than restating them, so it inherits the fix. Swept the Allowed divergence cell (`docs/contract-inventory.md:46`). Guards extended: the stamp mapping is asserted at each site that *selects* a trigger from a stamp (`tests/pr-review-contract.test.js:436-459`), while `templates/codex-workflow/prompts/fix-pr.md` repeats its cycle-1 trigger verbatim rather than selecting one, so it is asserted separately for the "already mapped upstream" and no-`@claude`-shorthand statements. Must-survive (1) is pinned in the pipeline too (`tests/milestone-pipeline.test.js:1135-1138,1174-1177`): a `[C5]` issue stamped fable posts the bare `@codex review effort:high`, never the C0–C10 row's luna and never `@codex fable`.

2. **`title` was routing-critical but unspecified in the prep contract** (Recommended Optional) — confirmed: `workflows/milestone-pipeline.js:703` reconciles against the title, the schema entry was a bare `title: { type: 'string' }`, and the prep prompt's extraction list never named the field. Both now require the title exactly as `gh issue view --json title` reports it, including the `[C<score>]` prefix, and say why (`workflows/milestone-pipeline.js:292`, `:681`). Fail-safe direction is unchanged; this removes the cost and the lost C0–C10 routing.

3. **`ex.effective_complexity` was written and never read** (Recommended Optional) — confirmed by repo-wide search: the only occurrence was the assignment itself, `ex` is never spread into a result record, and both review-routing readers moved to `review_complexity` earlier in this PR. Deleted the assignment (`workflows/milestone-pipeline.js:893`). The local `effectiveComplexity` still drives build re-routing and the rescore record, so the three "Must survive" cases hold: an unprefixed issue rescored to 5 stays on the top review band, a `[C5]` issue rescored to C90 still escalates its build band and records `from: 5, to: 90`, and a `missing_block` issue with no usable rescore still reaches `derivedBuild` as `undefined`.

Also corrected a stale comment in `tests/pr-review-contract.test.js:479-482`, which still described `@codex sonnet review` as matching no trigger after the previous cycle corrected that claim.

Verification: `bun test` — 258 pass, 0 fail.

**Plain simple English:** An operator can name the reviewer on an issue. When the run used Codex, two pages did not say how to translate that name, so they used the size band and gave the smallest change the cheapest reviewer. Both pages now carry the translation. A second fix tells the reading agent to copy the issue title word for word.

---

richkuo:
@claude opus review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-08-23T02:32:39Z) in cc20716.

### Fixed

1. **A stamped `@claude haiku review` reached the write-capable fix-pr route** — confirmed at HEAD: `workflows/milestone-pipeline.js:298` admits `haiku` in the `first_review_model` enum, `:176` maps it through `MODEL_IDS`, and the stamped branch of `firstReviewTrigger` emitted it verbatim. Traced through `.github/workflows/claude.yml`: `:144-149` has arms for `opus|opus5`, `sonnet|sonnet5`, `fable|fable5` only; `:253-255` takes the first token after `@claude` that is not one of those six as the route keyword, giving `KEYWORD=haiku`; `:266-271` then selects `MODE=fix-pr` on a trusted-author PR (`OWNER`/`MEMBER`/`COLLABORATOR`, or `claude[bot]` forced trusted at `:261`). Added `CLAUDE_REVIEW_SHORTHAND` (`workflows/milestone-pipeline.js:227-236`), the Claude column of the table `CODEX_REVIEW_SHORTHAND` already provides, mapping a stamped `haiku` onto `@claude sonnet review` and logging the normalization (`:267`). I chose the mapping over dropping `haiku` from the enum: `haiku` is a legal build model, the Codex column already maps it onto its cheap tier, and this PR writes it into four contract documents as a legitimate stamped reviewer — dropping it would discard the operator's intent instead of honoring it at the cheapest reviewer the Action admits. All three "Must survive" cases pass: a stamped `haiku` now posts `@claude sonnet review effort:high` at cycle 1 and at every blocking cycle (`tests/milestone-pipeline.test.js:1208-1213,1225`), the Codex route is unchanged, and an unstamped `[C5]` still posts `@claude sonnet review` while a stamped `@claude opus review` still repeats verbatim. Guard added at `tests/pr-review-contract.test.js:487-533`: it parses each Action's own route-keyword set out of the workflow and asserts every shorthand either review column can emit is admitted, plus every band model, plus that every `MODEL_IDS` build model is representable in both columns. I mutation-checked it — reverting the mapping fails both the new contract guard and the runtime case.

2. **`rereview-routing.md` let a band row override a non-Fable cycle-1 reviewer** — confirmed: `:19` stated the correct rule while `:17`'s heading ordered band routing after the cycle-1 check, the table at `:23-28` carried no precondition, and row `:28` asserted `@claude fable review effort:high` ran cycle 1, which a stamp can falsify at any score. Retitled the heading to route on the cycle-1 reviewer, added an explicit precondition that the band table applies **only** when the PR carries no `@<bot> … review` trigger comment at all, replaced the C81+ row's false assertion with `@claude opus review` and the reason a re-review never opens a Fable cycle, scoped the ladder to a Fable cycle 1 with the other-model case stated beside it, and corrected the score sources so the stamped `PR review:` line is no longer listed as one — it carries a trigger and no score. The command block no longer labels its lines with bands. Guard added at `tests/pr-review-contract.test.js:535-556`, also mutation-checked. All three "Must survive" cases now read consistently with `workflows/milestone-pipeline.js:280-283` and `:645`, and with `templates/claude-workflow/prompts/fix-pr.md:17`.

3. **The prep guard discarded a prefix it had already parsed** (Recommended Optional) — confirmed: the reconciliation computed `prefixScore` and used it only when a reported score disagreed, so an omitted field on a `[C50]` title left the run unscored on the top band. Added the mirror branch (`workflows/milestone-pipeline.js:712-720`): when the field is absent and the prefix is present, route on the prefix, with its own distinct log line. "Must survive" cases pinned at `tests/milestone-pipeline.test.js:397-400,421-425`: `[C50]` with the field omitted validates on Opus @ xhigh and posts `@claude opus review`; `[C50]` with a reported `0` still routes as unknown naming `[C50]`; an unprefixed title and `[Cx]` both stay unknown with no reconciliation log — the existing count assertion on that log still holds at 3.

Verification: `bun test` — 260 pass, 0 fail.

**Plain simple English:** An operator can name Haiku as the reviewer. The Claude job does not know that word, so it started the job that writes to the branch instead of the job that reviews. Haiku now becomes Sonnet, and a new test checks that every phrase the code can send is one the job accepts.

---

richkuo:
@claude opus review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-08-23T03:54:35Z) in 7bff45d.

### Fixed

1. **The Claude first-review prompt honoured a stamped model verbatim** — confirmed at HEAD: `templates/claude-workflow/prompts/issue-workflow.md:7` stated the stamped override with no model mapping, while its Codex sibling at `templates/codex-workflow/prompts/issue-workflow.md:11` carries one. Traced the consequence through `.github/workflows/claude.yml`: `:144-148` admits only `opus|opus5`, `sonnet|sonnet5`, `fable|fable5`; `:253-255` takes the first token after `@claude` that is not one of those six as the route keyword; `:266-271` then selects `MODE=fix-pr` on a trusted-author PR, and `:261` forces `claude[bot]` trusted — which is the author of a PR this Action opened. The prompt now maps the stamp before posting: `sonnet` or `haiku` to `@claude sonnet review`, `opus` to `@claude opus review`, `fable` to `@claude fable review`, each keeping the stamped `effort:<tier>`, with the route-keyword reason stated. Swept the two sites that author or render a stamp without naming a legal set — `skills/prd-to-issues/SKILL.md:78` now names `sonnet`/`opus`/`fable` and says never to stamp `haiku`; `skills/milestoneplan/SKILL.md:41` now renders a stamped `haiku` as `@claude sonnet review`, the trigger the runtime actually posts. Extended the override guard to require the mapping and to forbid a posting form of `@claude haiku` (`tests/pr-review-contract.test.js:664-686`), and pointed the shorthand guard at `templates/claude-workflow/workflows/claude.yml` and `templates/codex-workflow/workflows/codex.yml` as well (`:501-517`) — those are the copies a consumer repo installs. All three "Must survive" cases hold.

2. **The routing page never said which trigger comment is cycle 1** — confirmed: `skills/fix-pr-review/rereview-routing.md:19` said only "look at the PR's own trigger comments", and a search of `skills/` found no "earliest" rule anywhere, while all three siblings order it (`templates/claude-workflow/prompts/fix-pr.md:17`, `templates/codex-workflow/prompts/fix-pr.md:19`, `workflows/milestone-pipeline.js:508`). The scenario is real and reachable through this page's own text: `:58` has a non-blocking cycle post `@claude sonnet review`, which a newest-comment read would take for a non-Fable cycle 1 and then repeat on a C90 PR's blocking findings. `:19` now orders the chronological read and names the **EARLIEST** `@<bot> … review` comment as cycle 1, with that failure spelled out; the rung rule now counts from that earliest fable comment and says to ignore a `@claude sonnet review` while counting, since a non-blocking re-trigger consumes no rung. Guard at `tests/pr-review-contract.test.js:556-575`, which also pins the three siblings so they stay in step.

3. **A rescore crossing only a review boundary never reached review routing** (Recommended Optional) — confirmed: the escalation at `workflows/milestone-pipeline.js:905` gates on the **build** band index, and the review boundary at 10/11 sits inside build band `10–20`, so a `[C10]` rescored to 15 kept `review_complexity = 10` and reviewed on sonnet, against `skills/validate-issue/SKILL.md:100`. Review routing now compares the `REVIEW_BANDS` index directly and logs the escalation (`workflows/milestone-pipeline.js:916-931`). "Must survive" cases pinned at `tests/milestone-pipeline.test.js:1308-1335`: `[C10]` rescored to 15 posts `@claude review`; rescored to 8 stays `@claude sonnet review` with no escalation log; an unprefixed title rescored to 5 still takes the top review band (unchanged assertion).

4. **Step 10's pointer still called the ladder a C81+ rule** (Recommended Optional) — confirmed at `skills/fix-pr-review/SKILL.md:164`. Reworded to name the cycle-1 identification rule, the ladder and its floor, and the band table's precondition, and to state that the step-down keys to the reviewer that actually ran cycle 1 at any score. Added the file to the band-keyed-wording guard's consumer list (`tests/pr-review-contract.test.js:609`).

Verification: `bun test` — 261 pass, 0 fail. I mutation-checked each of the four guards: reverting any one fix fails its guard.

**Plain simple English:** An operator can name Haiku as the reviewer. The Claude job does not know that word, so it starts the job that writes to the branch. The prompt file now translates Haiku to Sonnet, as the pipeline already does. The routing page now says to take the oldest trigger comment, so a large change cannot get the cheapest reviewer.

---

richkuo:
@claude opus review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-08-23T04:10:21Z) in 046e41c.

### Fixed

1. **`work-on-issue-loop` step 1 stated the stamp override with no model mapping** — `skills/work-on-issue-loop/SKILL.md:30`. Confirmed: the paragraph restated the whole Claude band column and ended "a stamped `PR review:` line in its Execution block, which overrides the band", and the `:32` pointer at `fix-pr-review-loop` is framed as bot selection and preflight, so a Claude-default run never loads the mapping at `skills/fix-pr-review-loop/SKILL.md:45`. Traced verbatim `@claude haiku review effort:high` through `.github/workflows/claude.yml`: `:144-148` has arms for `opus|opus5`, `sonnet|sonnet5`, `fable|fable5` only; `:253-256` takes the first token after `@claude` that is none of those six as the route keyword, giving `KEYWORD=haiku`; `:259-264` marks an OWNER/MEMBER/COLLABORATOR author trusted and `:271` then sets `MODE=fix-pr`. `haiku` is an admitted stamp at `workflows/milestone-pipeline.js:312`. Added the clause the siblings carry — a stamped `sonnet` or `haiku` posts `@claude sonnet review`, `opus` posts `@claude opus review`, `fable` posts `@claude fable review`, each keeping the stamped `effort:<tier>` — plus the Codex column, and a dedicated guard in `tests/pr-review-contract.test.js` asserting the mapping, the admitted set, the route-keyword consequence, and that the file never writes `@claude haiku`.

### Resolved judgment calls

1. **Build vs review clamp asymmetry (Recommended Optional 2) — clamped the build, rather than documenting a carve-out.** Confirmed the asymmetry: `workflows/milestone-pipeline.js:933` keeps an unscored issue on the unknown/top band for review whatever the validator returns, while `:970-975` passed `effectiveComplexity` — which for an unscored issue is the raw rescore — to `derivedBuild`, so a rescore of 5 gave Sonnet 5 @ high with no plan where unknown gives Opus 5 @ xhigh with fableplan. Chose the clamp because the builder writes the code, so the argument that justifies the review clamp applies with more force, and because the pipeline already treats unknown as top band at every other stage. Passing `undefined` **is** the upward clamp — unknown is already the top band, so nothing can raise it. A missing-block issue whose title carries a score still derives from that score, and the upward escalation at `:950` still applies. Rejected alternative: documenting a build carve-out in `skills/validate-issue/complexity-scoring.md` — it keeps a cheaper default but leaves the strongest-blast-radius stage as the one stage that trusts a downward rescore. `complexity-scoring.md:40` and `:42` now state the rule for validate, build and review with no carve-out.

### Also fixed

1. **The missing-Execution-block test scored its fixtures `[C0]` (Recommended Optional 1)** — `tests/milestone-pipeline.test.js:1339-1342`. Confirmed: with `complexity: 0` agreeing with a `[C0]` title, `bandFor(0)` is band `0–9`, so for `#2` (rescore 6) and `#4` (rescore 9) the escalation at `:950` never fired and `derivedBuild(0)` produced the asserted `sonnet/high`. Restored unprefixed titles with `complexity` omitted, matching the schema at `:307`, and added scored missing-block fixtures (`[C6]`, `[C10]`, `[C5]` rescored 90) so the derivation-from-score path and the 9/10 boundary stay covered. The real-zero case remains covered by the separate test at `:374`. Corrected the stale comment.

**Verification:** `bun test` — 261 pass, 0 fail. Both new guards mutation-checked: removing the mapping clause fails the contract guard, and reverting the build clamp fails the pipeline test.

---

richkuo:
@claude opus review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-08-23T04:31:47Z) in 18a7d49.

### Fixed

1. **The cycle-1 read took the earliest trigger comment without excluding the cheap non-blocking re-trigger** — `skills/fix-pr-review/rereview-routing.md:19` and `:25`, `templates/claude-workflow/prompts/fix-pr.md:17`, `templates/codex-workflow/prompts/fix-pr.md:19`. Confirmed the reachable path: `:58` has a non-blocking pass post `@claude sonnet review` / `@codex luna review` at any band, `:54` excluded that comment only while counting ladder rungs, and `skills/fix-pr-review-loop/SKILL.md:25` treats already-present feedback as review 1 and posts no trigger — so on a C90 PR the earliest trigger comment can be that cheap one, `:19` then reads it as cycle 1, and `:25` also disables the band fallback. All four sites now skip the cheap shorthand during the cycle-1 read unless the PR's band is C0–C10, the one band that selects the same phrase as a first review, and the band table applies when the skip leaves no trigger comment. `workflows/milestone-pipeline.js:508` tells its agent to confirm the computed trigger against the earliest comment, so it now names the same skip. Guards added in `tests/pr-review-contract.test.js` assert the skip clause, the widened band gate, and the same pair on both Action prompts.

2. **A validator rescore deleted a stamped `PR review:` trigger even when the new band's reviewer was weaker** — `workflows/milestone-pipeline.js:950` and `:963-964`. Confirmed: the branch fires on an upward BUILD-band move, and the review scale is independent, so a `[C5]` issue stamping `@claude fable review effort:high` rescored to 25 crossed `bandFor` index 0 → 2, lost the stamp, and fell to `reviewBandFor(25).review.model === null`, the bare standard trigger — weaker, on an issue just judged harder. `runSubagentReviewLoop` reads the same fields, so subagent mode degraded the same way. Added `reviewModelRank` over the `REVIEW_BANDS` order (a stamped `haiku` ranks where sonnet does, since that is the trigger it posts) and made the replacement conditional on the rescored band's reviewer outranking the stamp. Build, plan and fableplan stamps keep their unconditional upward replacement. Both outcomes are logged. New runtime cases cover a kept Fable stamp that still steps down to opus, a replaced sonnet stamp on a real escalation to C90, a kept opus stamp, and the Codex column keeping `@codex review effort:high`.

**Verification:** `bun test` — 263 pass, 0 fail. Both fixes mutation-checked: removing the skip clause fails the routing-page guard, and making the stamp deletion unconditional fails both new pipeline tests.

**Note on this cycle's re-review trigger:** the maintainer directed `@claude review` for this cycle, which overrides the step-10 routing that would otherwise repeat cycle 1's `@claude opus review`.

---

richkuo:
@claude review

richkuo:
@claude review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-08-23T05:04:06Z) in dab9a4a. Verdict was `LGTM`; both `Recommended Optional` items are implemented.

### Fixed

1. **The stamped-first-review re-evaluation was gated on a build-band increase** — `workflows/milestone-pipeline.js:961` (the `BANDS.indexOf(...)` branch) with the stamp block nested inside it. Confirmed: the review boundary at 10/11 sits inside build band `10–20`, so a `[C10]` issue stamped `@claude sonnet review` rescored to 15 raised `ex.review_complexity` to 15 through the review clamp but never entered the block, and `firstReviewTrigger` then posted `@claude sonnet review` while `reviewModelRank(null)` is 1 against `reviewModelRank('sonnet')` 0 — the retained stamp ranked below the band default the same code computes. Moved the block out of the build branch to sit immediately after `ex.review_complexity` is assigned, and keyed its gate to `REVIEW_BANDS.indexOf(reviewBandFor(ex.review_complexity)) > REVIEW_BANDS.indexOf(reviewBandFor(ex.complexity))`. The replace-only-when-it-outranks rule inside is unchanged. An issue with no rescore never raises its review band, so it never enters the block and a stamped reviewer still survives every cycle at its own band. New cases cover all four shapes: `[C10]` stamped sonnet rescored 15 takes the bare trigger; `[C10]` stamped opus rescored 15 keeps opus; `[C10]` stamped sonnet rescored 8 keeps sonnet with no re-evaluation logged; an unstamped `[C10]` rescored 15 still escalates.

2. **`skills/milestone-workflow/SKILL.md:59` claimed the rescore replaces the review stamp unconditionally** — confirmed, and accurate only before the previous commit. The line now states the two rules separately: build model, build effort and fableplan always move to the escalated band's defaults, while a stamped `PR review:` reviewer moves only when the escalated review band's default outranks it. Added a guard in `tests/pr-review-contract.test.js` asserting both halves and refusing the old wording.

**Verification:** `bun test` — 265 pass, 0 fail. Both fixes mutation-checked: reverting the gate to the build index fails the new boundary test, and restoring the old doc sentence fails the new contract guard.

---

richkuo:
@claude sonnet review

PR #183, opened 2026-08-23
docs(#180): add a worked example review to the pr-review skill [C35, Opus 5, high]

### LINKED ISSUE #180: [C35] The PR review skill gives no example review comment, so the output format drifts
**Complexity: 35/100** — Capability 1 (Uncertainty 2 — the example must cover every field and stay correct as rules change); Volume 10 — Opus 5, high · fableplan: no

## Problem

`skills/pr-review/SKILL.md` defines a strict output shape and shows none of it. The `## Format` section (lines 28-42) specifies four H3 finding sections with different blocking behavior, a first line that is exactly `LGTM` or `Needs Updates`, numbered items with a bold title and a description, and up to four extra fields whose order is fixed: `Invariant:` and `Must survive:` on two sections, `Recommended proposed solution:` on one, and `Plain simple English:` last on every finding (line 40). A `**Verification limitation:**` line has its own placement rule and must never sit under an H3 section (line 34).

That is fifteen rules of prose describing a document layout, with no rendering of the layout. The reviewer must build the shape from the description on every run.

The cost lands on the fixer. `skills/fix-pr-review/SKILL.md` step 3 parses reviews back into four buckets by section heading, and treats a `**Verification limitation:**` line as a non-finding that must be skipped when classifying. Format drift on the reviewer side turns into misclassification on the fixer side.

`skills/pr-review/` holds only `SKILL.md`. There is no reference file and no template.

## Goal

A reviewer can see one correct review comment end to end before it writes, and match its output against it. The example carries the field order and section behavior that the prose describes.

## Approach

Add a worked example to `skills/pr-review/`. Put it in a separate reference file rather than inline, so the skill body stays scannable and the example can be long enough to be useful.

Contents of the example:

1. One complete `Needs Updates` review over an invented small diff, showing all four H3 sections in order, with one item each. Each item shows its full field set in the required order, so `Invariant:` / `Must survive:` / `Recommended proposed solution:` / `Plain simple English:` appear where the rules put them.
2. One `**Verification limitation:**` line placed correctly, between the last finding section and the footer.
3. One bare `LGTM` review, showing that the verdict stands alone above the footer with nothing else.
4. The attribution footer with verb `Validated`, per line 42.

The example diff should be a small agent-instruction file, because line 15 makes those executable and reviewers under-review them. That gives the example a second job: it demonstrates reviewing a skill or workflow file for behavioral defects rather than prose quality.

Add one line in `## Format` pointing at the file, phrased so the prose stays authoritative. The example illustrates the rules; where they disagree, the prose wins. Say that explicitly, so a future rule change that outruns the example does not silently redefine the format.

Touch-set:

- `skills/pr-review/example-review.md` — new file with the four items above.
- `skills/pr-review/SKILL.md` — one pointer line in `## Format`, with the precedence statement.
- `README.md` — check whether it lists per-skill files; add the new file if it does.

## Acceptance criteria

- `skills/pr-review/` contains a reference file with a complete `Needs Updates` review showing all four H3 sections, each with one fully-fielded item in the required field order.
- The same file shows a correctly placed `**Verification limitation:**` line and a standalone `LGTM` review.
- Both examples end with the attribution footer using verb `Validated`.
- `skills/pr-review/SKILL.md` points at the file and states that the prose rules win over the example where the two disagree.
- Every field name and section heading in the example matches the current `## Format` rules exactly.

## Plain simple English

The review rules describe a strict layout in words, with no example of a finished review. Each reviewer must build the layout from the description, and small differences appear. The tool that reads reviews then sorts the results into the wrong groups. This change adds one correct example review for the reviewer to copy.

---

richkuo:
## Summary

`skills/pr-review/SKILL.md` specified a strict review layout in prose and rendered none of it, so every reviewer rebuilt the shape from the description. `skills/fix-pr-review/SKILL.md` step 3 parses reviews back into four buckets by section heading, so reviewer-side drift becomes fixer-side misclassification.

- **New `skills/pr-review/example-review.md`** — a worked example with:
  - one complete `Needs Updates` review over an invented agent-instruction diff (a `release-notes` skill file), showing all four H3 sections in blocking-first order (`### Needs Fixing`, `### Requires Human Review`, `### Recommended Optional`, `### Create Follow-up Issue`), one item each, every item carrying its full field set in the required order (`Invariant:` / `Must survive:` / `Recommended proposed solution:` where they apply, `Plain simple English:` always last);
  - one `**Verification limitation:**` line in the correct place — after the last finding section, above the footer, with no finding fields and outside every H3 section;
  - one bare `LGTM` review, verdict alone above the footer;
  - the LLM Attribution Footer with the verb `Validated` on both examples.
- **`skills/pr-review/SKILL.md`** — one pointer bullet at the end of `## Format`, stating that the prose rules win and the example is stale where the two disagree.
- **New `tests/pr-review-example.test.js`** — nine tests that make each acceptance criterion script-verifiable and stop the example from drifting away from the rules it illustrates.

- **The example lives in a fenced block, with commentary outside it.** The posted-review text stays copyable and pure; the "what this demonstrates" notes sit between blocks and are explicitly marked as never part of a review.
- **The invented diff is an agent-instruction file**, per the issue: the example doubles as a demonstration of reviewing a skill file for behavioral defects and self-consistency instead of prose quality. Its findings are a step-to-step contradiction, an unconfirmed-publish policy question, an undefined input class, and a cross-skill tag mismatch.
- **Added a guard test rather than relying on review.** The repo guards its duplicated contracts with `bun test` (`tests/pr-review-contract.test.js` and friends), and the example is a rendered copy of the format rules. The test asserts section order, per-section field sets and their order, that every field name and section heading the example uses also appears in `SKILL.md`, the limitation line's placement and shape, the standalone `LGTM` shape, both `Validated` footers, and the word caps (under 55 words for plain-simple-English and recommended-solution fields, under 50 for the `Requires Human Review` technical description). It is a new file rather than an append to `tests/pr-review-contract.test.js`, which is scoped to the five verification-method contract copies.
- **`README.md` left unchanged.** The issue's touch-set said to add the file "if README lists per-skill files". It does not — `## Skills` is a table of skills with descriptions, and no reference file of `fix-pr-review` or `validate-issue` is listed there either. The skill body is the discovery path.
- **`docs/contract-inventory.md` left unchanged.** That inventory declares itself scoped to the loop/validate pipeline rules and explicitly excludes broader duplicated-contract tracking. The new guard test covers the drift risk.
- **The Action prompt copies were left unchanged.** `templates/claude-workflow/prompts/pr-review-format.md` and `templates/codex-workflow/prompts/pr-review-format.md` are standalone prompts on a runner that has no copy of the skill directory, so a pointer to `example-review.md` would dangle there. Inlining a full example into those prompts is a separate scope decision.
- **Footer model line.** The dispatch instruction named `Fable 5`; the model that actually authored this change is Opus 5, so the footer names Opus 5.

## Verification (measured)

- Baseline before any change: `bun test` → **249 pass, 0 fail, 1738 expect() calls, 16 files**.
- After the change: `bun test` → **258 pass, 0 fail, 1802 expect() calls, 17 files**. Zero new failures; the nine added tests are the delta.
- `bun test tests/pr-review-example.test.js` → **9 pass, 0 fail**.
- Acceptance criteria, each covered by a named test in `tests/pr-review-example.test.js`: four sections in order with one fully-fielded item each; correctly placed `**Verification limitation:**` line; standalone `LGTM`; `Validated` footer on both examples; the `SKILL.md` pointer plus the precedence statement; every field name and section heading in the example present in the current `## Format` rules.
- Word caps measured by the tests, not by eye: the longest plain-simple-English field is 45 words, the `Recommended proposed solution:` field is 48 words, and the `Requires Human Review` technical description is 49 words.

Nothing in this change runs on a device or calls a live API, so there is nothing pending manual verification. The example's review findings describe an invented file; they are illustrative and were never run against real `gh` commands, which the example's own text states.

Closes #180

## Plain simple English

The rules for a pull request review describe a strict layout in words only. Each reviewer must build that layout again from the description, and small differences appear. The tool that reads reviews then puts findings in the wrong groups. This change adds one correct example review to copy, and a test that keeps the example equal to the rules.

---


richkuo:
@claude review

---

PR #185, opened 2026-08-23
feat(#178): make a review read the prior cycles before it drafts [C37, Opus 5, high]

### LINKED ISSUE #178: [C37] The PR review skill does not read prior review cycles, so a refuted finding can come back
**Complexity: 37/100** — Capability 1 (Risk 2 — review-gate process, reversible, no money/data path); Volume 12 — Opus 5, high · fableplan: no

## Problem

`skills/pr-review/SKILL.md` never tells the reviewer to read what happened on earlier review cycles of the same pull request. The "Before you write — verification method" list (`skills/pr-review/SKILL.md:12-17`) sources the review from the diff, the changed files, and external primary sources. It says nothing about the prior review comments or the fixer's disposition replies that are already on the pull request.

Those replies exist and carry evidence. `skills/fix-pr-review/SKILL.md:155` (step 9) posts a per-finding disposition comment, and step 4 requires a code-grounded rebuttal for every finding it marks Refuted. The fixer then ignores its own prior disposition comments when it collects the next round of feedback (`skills/fix-pr-review/SKILL.md:46`), which is correct for the fixer — those are its own words.

The result is that nothing in the loop carries a refutation forward. A re-review can raise the same finding again, with no obligation to address the rebuttal that answered it. The fixer must then re-validate the same claim from scratch and write the same rebuttal again. `skills/fix-pr-review-loop/SKILL.md:78` (step 3 rule 3) states that a `Needs Updates` verdict continues the loop at any cycle count — only an `LGTM` stops it — so a finding that keeps returning keeps the pull request open indefinitely.

The 5-cycle relaxation in `skills/fix-pr-review-loop/SKILL.md:77` limits the damage for non-blocking findings only. It does not apply to a repeated blocking finding.

## Goal

A reviewer reads the prior cycles before it writes. A finding that a previous cycle refuted with code evidence comes back only when the reviewer can say why that evidence fails. Real defects still get re-raised; settled ones stop consuming cycles.

## Approach

Correctness and safety outrank diff size here: the rule must never become a way to suppress a real defect.

Add one bullet to the "Before you write — verification method" list in `skills/pr-review/SKILL.md`, after the self-consistency bullet at line 13:

- **Read the prior cycles before you write.** Fetch the earlier review comments and the fixer's disposition replies on this pull request. For every finding you are about to raise that a prior cycle recorded as Refuted or Partial with a code-grounded rebuttal, you may re-raise it only by naming that rebuttal and stating, from current code at `file:line`, why it does not hold — a changed code path, a case the rebuttal did not cover, or a factual error in it. A finding re-raised without that treatment is dropped.

Two carve-outs belong in the same bullet:

1. **Safety carve-out wins.** The existing safety carve-out (`skills/pr-review/SKILL.md:31`) already overrides materiality and confidence. State that it also overrides this rule: a money, data-integrity, security, or auto-protective finding is always surfaced, and when the prior rebuttal is the only reason to drop it and you cannot confirm the rebuttal from code, it goes under `### Requires Human Review` rather than being dropped.
2. **A rebuttal is not a shield for new evidence.** A prior refutation of finding X does not settle a different defect in the same file or function. The rule matches findings by claim, not by location.

Touch-set:

- `skills/pr-review/SKILL.md` — the new bullet; a matching clause in the `LGTM precondition` bullet (line 35) so that failing to fetch the prior cycles is an incomplete applicable item, the same as any other verification step.
- `skills/fix-pr-review/SKILL.md` — no behavior change, but check that step 9's disposition format states the verdict and rebuttal clearly enough for a later reviewer to match a finding to it. If it does not, make the per-finding disposition line carry the finding's claim verbatim.
- `skills/fix-pr-review-loop/SKILL.md` and `skills/work-on-issue-loop/SKILL.md` — check whether either needs a note that the reviewer now reads prior dispositions. Add one only if a step there would otherwise contradict it.

Follow the repo convention of putting the rule in the skill that owns the behavior. `pr-review` owns what a reviewer must do before writing, so the rule lives there.

## Acceptance criteria

- `skills/pr-review/SKILL.md` requires the reviewer to fetch prior review comments and disposition replies before drafting, and names them as a source alongside the diff.
- The skill states the bar for re-raising a previously refuted finding: name the rebuttal and show from current code why it fails.
- The skill states that the safety carve-out overrides the new rule, and routes an unconfirmable safety-class finding to `### Requires Human Review` instead of dropping it.
- The skill states that the rule matches findings by claim, so a prior rebuttal does not cover a different defect in the same file.
- The `LGTM precondition` bullet counts the prior-cycle read as an applicable item.
- A search of `skills/` finds no step in `fix-pr-review`, `fix-pr-review-loop`, or `work-on-issue-loop` that contradicts the new rule.

## Plain simple English

When a review finds a fault, the person who fixes the pull request can answer that the fault is not real, and show why. The next review does not read that answer. It can raise the same point again, and the cycle repeats. This change makes each review read the earlier answers first, and reopen a settled point only with new evidence.

---

richkuo:
## Summary

A pull request review now reads what earlier cycles decided before it drafts. A finding that a previous cycle refuted with code evidence returns only when the reviewer names that rebuttal and shows, from current code at `file:line`, why it fails. A re-raise without that treatment is dropped, so a settled point stops consuming review cycles.

Two carve-outs bound the rule so it can never suppress a real defect:

1. The safety carve-out wins. A money, data-integrity, security, or auto-protective finding is always surfaced, and when a prior rebuttal is the only reason to drop it and the rebuttal cannot be confirmed from current code, it goes under `### Requires Human Review`. It is never dropped.
2. Findings match by claim. A prior rebuttal settles only the claim it answered, so a different defect in the same file, function, or line is raised normally.

A route that cannot read the pull request's comments emits a non-blocking `**Verification limitation:** prior review cycles unreadable — <access reason>.` line and reviews from the diff alone. That mirrors the existing unreachable-primary-source handling and keeps an autonomous loop from stalling on a harness property.

## Verification

Measured in this worktree with `bun test` (Bun 1.4.0):

- Baseline before any edit: **249 pass, 0 fail**, 1738 expect() calls, 16 files.
- After the change: **254 pass, 0 fail**, 1854 expect() calls, 16 files. Five new tests, zero new failures.
- `bun test tests/pr-review-contract.test.js`: 22 pass, 0 fail, 419 expect() calls.

Other checks actually run:

- All four workflow YAML files parse (`templates/claude-review.yml`, `templates/codex-review.yml`, `.github/workflows/claude-run.yml`, `.github/workflows/codex-run.yml`) through the `yaml` parser, and the staged run block round-trips out of the block scalar as valid shell.
- The staging `jq` filter was run against a real pull request (`gh pr view 175 --repo richkuo/rk-skills --json comments,reviews --jq ...`) and produced the expected per-comment blocks, which confirms the field names.
- The staging tail was simulated under `bash -e` with a stub `gh` in three modes: entries returned (appended under the header), no entries (header only, which reads as the first cycle), and a failed fetch (file replaced with the unavailable marker, job still exits 0).
- The prompt files still contain zero double quotes, backticks, or dollar signs, which `tests/prompt-shell-safety.test.js` enforces.

Acceptance criteria are grep-verifiable through `tests/pr-review-contract.test.js`: `makes the reviewer read the prior cycles before it drafts` asserts every clause across all five contract copies, `dispositions carry the finding claim verbatim so a later review can match it` covers the fixer side, and `no fixer or loop step contradicts the reviewer reading prior cycles` covers the last criterion.

- **The rule lives in `pr-review`**, which owns what a reviewer must do before writing, and is mirrored into the four other review-contract copies that `tests/pr-review-contract.test.js` already keeps in sync.
- **The `LGTM precondition` bullet now names the prior-cycle read** as an applicable item, so skipping it is an incomplete item like any other verification step.
- **The Codex review route needed more than prose.** It runs in a read-only sandbox with no network, so `gh` cannot reach GitHub and the rule would have been inert there. Its workflow now stages the earlier reviews, review-formatted comments, and disposition replies to `.rk-prior-review-cycles.md` in the workspace, beside the diff it already stages, and the prompt marks that file as untrusted data. An empty entry list means the first cycle; only a failed fetch marks the file unavailable. This goes beyond the issue's touch-set, and the alternative was a rule that always degrades to a verification limitation on that route.
- **The fixer keeps ignoring its own dispositions** when it collects new work. That scoping is now stated as the fixer's alone, with the reason the reviewer reads those same comments, plus a rule against deleting, editing, or burying a disposition comment.
- **Disposition titles are copied verbatim.** The next reviewer matches findings to dispositions by claim, so `skills/fix-pr-review/disposition-comment.md` and both Action `fix-pr` prompts now require the reviewer's own bold title word for word.
- **`docs/contract-inventory.md` gains a row** for the new five-copy rule, following the repo convention that a replicated contract rule is inventoried with its guard.

## Pending verification

Not verifiable in this headless environment, and not claimed as verified:

- No GitHub Actions run exercised the edited workflows. The staged-file read by the Codex agent inside the read-only sandbox is unconfirmed; the staging step itself was simulated locally and the YAML parses.
- No live review was performed under the new prompt, so the behavioral effect on a real re-review cycle is unmeasured.

Closes #178

## Plain simple English

When a review finds a fault, the person who fixes the pull request can show that the fault is not real. Before, the next review did not read that answer, and it could make the same point again. Now each review reads the earlier answers first. It can only open a settled point again with new proof from the code. Safety faults are always shown.

---


richkuo:
@claude review

---

richkuo:
@claude opus review

PR #193, opened 2026-08-23
feat(pr-review): gate review remedies on scope and brake a diverging loop [C66, Opus 5, high]

richkuo:
## Summary

The review chain has no bound on how much a PR may grow while answering reviews. Three rules combine to allow it:

- **`fix-pr-review` step 6** implements every Confirmed, Judgment and `Recommended Optional` finding at the absolute-best standard, with no test for whether the remedy belongs in this PR.
- **`pr-review`** routes `### Create Follow-up Issue` as "the disposition of last resort — prefer keeping work in the PR", which pushes a new-mechanism remedy into `### Recommended Optional` instead.
- **`fix-pr-review-loop` step 3** states outright that `Needs Updates` never stops the loop by cycle count, so a PR that keeps generating findings in its own new code loops until an LGTM appears.

Each rule is defensible alone. Together they have no upper bound.

### The case this comes from

One PR ran 11 review cycles. Every finding was correct, and almost every one was a defect in code the *previous* cycle had just added. A two-line fix reached 1304 added lines across two subsystems the issue never asked for — a foreign-cookie sweeper and a persisted revocation journal. The journal entered as a single `Recommended Optional` suggestion in cycle 7 and produced four further cycles on its own.

The replacement PR, built to the same correctness bar with both subsystems dropped, was 429 lines and converged in three cycles.

## What changes

**1. A scope test in `fix-pr-review` step 4**, decided per finding alongside the verdict.

| Class | Test | Action |
|---|---|---|
| In scope | The remedy corrects code this PR already adds or changes | Implement it |
| Out of scope | The remedy introduces a mechanism the PR does not have — new persistent store, lifecycle or generation scheme, cross-cutting invariant, retry or recovery path, subsystem — **and** the issue does not ask for it | File an issue; do not implement |

Out-of-scope findings are filed, never implemented and never silently dropped. **The source of the defect decides scope, not the size of the remedy**, so "it is only a few lines" cannot reopen the gate, and a defect in code the PR added stays in the PR however much mechanism its fix needs. A `Recommended Optional` item is a suggestion and carries no authority to enlarge the PR. The safety carve-out overrides the out-of-scope class in one direction only: where **this PR** creates the hazard.

Step 6 gains the matching stop: if implementation reveals that an in-scope remedy cannot work without a new mechanism, that is a scope-test failure rather than a licence to build it.

**2. A reviewer-side counterpart in `pr-review`.** A remedy the PR has no mechanism for meets the existing "own design decision" bar on its own, so it belongs under `### Create Follow-up Issue` rather than `### Recommended Optional`.

**3. A divergence brake in `fix-pr-review-loop` step 3** — `review_count >= 4`, verdict `Needs Updates`, and every blocking finding sitting in code an earlier cycle added. Both halves are required: a PR still being corrected in its *original* work keeps going however many cycles it takes. The brake stops and hands the scope decision to the user with the cycle chain and the cumulative diff, rather than trimming scope on its own.

Both Action `fix-pr` prompts restate the scope test, so the workflow fix path does not keep the unbounded behavior. `docs/contract-inventory.md` gains the scope-test row, and its review-cycle row records the brake as the one exception to "`Needs Updates` alone never stops by cycle count".

## Test change, called out explicitly

`tests/loop-validate-pipeline-contract.test.js` asserted the old invariant verbatim:

```
/Needs Updates.*never stops|never force-stops a `Needs Updates`|never stops the loop by cycle count/is
```

That is the design this PR intentionally changes, so the assertion is updated rather than deleted, and the replacement is strictly more specific: a cycle count alone still never stops a `Needs Updates` PR, and the brake must carry its second condition to satisfy the new test.

## Verification

289 pass, 0 fail. Thirteen mutations were run, each turning the suite red on its own with the clean tree green:

| Mutation | Result |
|---|---|
| brake loses its self-inflicted condition (`fix-pr-review-loop`) | 1 fail |
| brake removed entirely | 1 fail |
| brake threshold raised above the LGTM cap | 1 fail |
| brake loses its self-inflicted condition (`work-on-issue-loop`) | 1 fail |
| `work-on-issue-loop` drops the threshold | 1 fail |
| scope test loses its keying rule | 1 fail |
| out-of-scope findings implemented anyway | 1 fail |
| step 6 stops honoring the scope test | 1 fail |
| `pr-review` new-mechanism rule removed | 1 fail |
| `pr-review` loses the routing test | 1 fail |
| Claude Action prompt loses the scope test | 1 fail |
| Codex Action prompt stops filing out-of-scope findings | 1 fail |
| inventory loses the scope-test row | 1 fail |
| clean tree | 0 fail |

The new assertions are scoped to the region that states each rule. An earlier draft searched the whole document and passed on two mutations, because the Red Flags and Common Mistakes rows restate the same phrases and satisfied the match independently.

## Plain simple English

When a reviewer finds a problem, the fixer builds whatever the fix needs, and the loop keeps going until the reviewer approves. On one pull request this turned a two-line fix into 1304 lines over 11 rounds. Now a fix that needs a new mechanism becomes its own issue, and a loop that keeps correcting its own corrections stops and asks the user.

---


richkuo:
@claude opus review

richkuo:

richkuo:
Addressed review feedback (Opus 5 review via github-actions[bot] · 2026-08-23T10:11:15Z) in 0fa49bf.

### Fixed
1. **Step 6's "a new mechanism is a scope-test failure" rule reverses step 4's source-of-defect rule and lets the PR defer a hazard it created itself.** — Step 6's discovery bullet now re-runs step 4's precedence: a rule-1 finding (PR-caused defect or hazard) stays in scope and gets the mechanism built here; only rule-3 findings move to rule 2 and get filed (`skills/fix-pr-review/SKILL.md:135`); both Action prompts carry the same precedence (`templates/claude-workflow/prompts/fix-pr.md:9`, `templates/codex-workflow/prompts/fix-pr.md:11`).
2. **The scope table leaves a cheap fix to a pre-existing defect unclassified, and the prose that fills the gap deadlocks the loop against `pr-review`'s routing.** — The scope test is now three rules applied in order; rule 3 puts every mechanism-free remedy in scope, including the pre-existing trivially-fixable instance the reviewer routes into the PR, so fixer and reviewer agree (`skills/fix-pr-review/SKILL.md:101-109`). The "gap that predates the PR is out of scope however easy the patch looks" sentence is gone; remedy size never decides scope in either direction.
3. **The reviewer-side rule reaches no reviewer that actually runs.** — The routing rule now appears in all four Action consumers (`templates/claude-workflow/prompts/pr-review-format.md:59`, `templates/codex-workflow/prompts/pr-review-format.md:59`, `templates/claude-review.yml:271`, `templates/codex-review.yml:324`); the inventory row lists them (`docs/contract-inventory.md:48`) and `tests/pr-review-contract.test.js` guards each consumer and the row.
4. **The scope test has no defined result when the pull request closes no issue — the case this pull request is itself in.** — The yardstick is now named: the closed issue(s)' asks (union when several), else the PR body's own stated scope (`skills/fix-pr-review/SKILL.md:99`, both fix-pr prompts, and fix-pr-review-loop's Diverging row now names the same yardstick).
5. **`skills/pr-review/SKILL.md:53` states two opposite routes for one finding.** — The bullet is rewritten as three rules applied in order with rule 1 (PR-caused, stays in PR) explicitly outranking rule 2 (new mechanism, files), and rule 3 preserving the trivially-fixable routing above it, so each finding has exactly one route (`skills/pr-review/SKILL.md:53`).
6. **The brake's escalation target is asserted over the whole document, so the mutation that removes it still passes.** — The Diverging assertion now reads the stop-condition region (`tests/loop-validate-pipeline-contract.test.js:195`).
7. **The growth check writes two numbers into two artifacts whose own specs have no slot for them.** — Step 11's content list names both numbers, and the disposition template gains a `Growth check:` line with an omit-when-quiet rule (`skills/fix-pr-review/SKILL.md:190`, `skills/fix-pr-review/disposition-comment.md:12,35`). This comment omits it: cycle 2, diff well under 3x.
8. **Nothing stops the fixer from filing the same out-of-scope issue on every cycle.** — Step 6 and both fix-pr prompts now require `gh issue list --search "<keywords>" --state all` before filing, citing an existing issue (and noting a human-closed one) instead of duplicating (`skills/fix-pr-review/SKILL.md:132`).
9. **Rule 3's second condition and the growth check both name inputs with no stated way to obtain them.** — `<first-push-sha>` derives from `gh pr view --json commits` with the cycle-1 trigger timestamp (PR `createdAt` fallback), the cycle count from the PR's trigger comments per rereview-routing, and cycle attribution from `git blame` plus `git merge-base --is-ancestor` against the first-push SHA — all read from the PR, so a resumed loop gets the same values (`skills/fix-pr-review/SKILL.md:113`, `skills/fix-pr-review-loop/SKILL.md:83`).

All 293 tests pass (`bun test`).

---

richkuo:
@claude opus review

richkuo:
Addressed review feedback (Opus 5 review via github-actions[bot] · 2026-08-23T19:38:47Z) in c66a648.

### Fixed
1. **A blocking finding the fixer defers is never fixed, never settled, and never braked, so the loop cannot converge.** — A `Deferred to follow-up` disposition is now a terminal disposition alongside `Refuted` and `Corrected scope (partial)`: it settles a finding when it names **both** the scope rule it applied and the issue it filed, and the next reviewer re-raises it only by naming the deferral and showing from current code that it fails — the remedy needs no mechanism the PR lacks, or scope rule 1 claims the defect, which no deferral outranks. A deferral missing either half settles nothing (`skills/pr-review/SKILL.md:14`, all four Action consumers, `skills/fix-pr-review/disposition-comment.md:26,35`, both fix-pr prompts, `docs/contract-inventory.md:51`).
2. **The growth check and the brake both count base-branch code that step 7 merged into the head.** — Both now exclude the base. The growth check compares the PR's own diff against the base at HEAD with the same reading taken at the first push, and forbids the plain two-dot diff by name (`skills/fix-pr-review/SKILL.md:113`). The self-inflicted test requires the blame commit to be an ancestor of neither `<first-push-sha>` nor `origin/<baseRefName>`, so a line a step 7 merge brought in is base-branch work (`skills/fix-pr-review-loop/SKILL.md:83`); the Diverging report uses the same base-excluded pair (`:103`).
3. **Rule 1 routes every finding in PR-added code away from `### Requires Human Review`.** — The ordered rules now route only between keeping a finding in this PR and filing it as a follow-up issue, and state that they never remove a finding's eligibility for `### Requires Human Review` — a real tradeoff, provably missing context, an unconfirmable safety finding, or an LGTM-precondition gap escalates whatever the rules say (`skills/pr-review/SKILL.md:53`, `templates/claude-workflow/prompts/pr-review-format.md:59`, `templates/codex-workflow/prompts/pr-review-format.md:59`, `templates/claude-review.yml:280`, `templates/codex-review.yml:333`).
4. **The scope table tells the fixer to implement findings the reviewer routed to `### Create Follow-up Issue`.** — The table now excludes a reviewer-routed follow-up item outright: it is filed and never implemented, whatever the rules would say. Scope rule 1 is the exclusion's single stated exception — a defect in code this PR adds or changes, or a hazard it creates, is fixed here however the reviewer routed it, and the disposition says so. Step 6, step 1's LGTM path and both fix-pr prompts carry the same one outcome (`skills/fix-pr-review/SKILL.md:53,102,130`, `templates/claude-workflow/prompts/fix-pr.md:7,9`, `templates/codex-workflow/prompts/fix-pr.md:9,11`).
5. **The reviewer-side rule 2 drops "a new subsystem" from the very list its own closing sentence calls the canonical case.** — All four Action consumers now carry the five-item list and the closing sentence, and the guard checks each list entry by name in the owner and in every consumer, so a later addition fails when one copy is missed (`templates/claude-workflow/prompts/pr-review-format.md:59`, `templates/codex-workflow/prompts/pr-review-format.md:59`, `templates/claude-review.yml:280`, `templates/codex-review.yml:333`, `tests/pr-review-contract.test.js:1140`).
6. **`review_count` names two different quantities inside the brake.** — The brake's derived count is now `pr_cycle_count`, stated as distinct from the in-memory `review_count` that rules 2 and 4 read, and every restatement carries that definition (`skills/fix-pr-review-loop/SKILL.md:79,83,121`, `skills/work-on-issue-loop/SKILL.md:43`, `skills/fix-pr-review/SKILL.md:113`, `docs/contract-inventory.md:46`).
7. **The brake's second condition has no defined result for a blocking finding with no citable line.** — An unattributable blocking finding — no `file:line`, or a path the head no longer has — now **defeats** the condition rather than satisfying it: rule 3 fires only when the review carries at least one blocking finding and every one of them is attributed to a later cycle (`skills/fix-pr-review-loop/SKILL.md:83`, restated at `skills/work-on-issue-loop/SKILL.md:43`).
8. **The worked example for the optional-findings rule teaches an outcome rule 1 forbids.** — The example now names its class: the durability suggestion was about **pre-existing** state, so it is a rule-2 remedy and gets filed; had the gap sat in code the PR added, rule 1 would have matched first and the journal would have been built here (`skills/fix-pr-review/SKILL.md:111`).
9. **The `Needs Updates` invariant is asserted over the whole document, so deleting it from rule 4 keeps the test green.** — The assertion moves into `STOP_CONDITION_REGION` and accepts rule 4's own "no cycle count alone stops a" clause, so a restatement in step 5's prose can no longer satisfy it (`tests/loop-validate-pipeline-contract.test.js:167-206`).

### Test edits
1. `tests/loop-validate-pipeline-contract.test.js` — the `Needs Updates` invariant assertion moved from the whole document into the stop-condition region. Case: **Wrong**; the expectation was never correct at document scope. Ground: the region discipline stated in the same file's comment at `tests/loop-validate-pipeline-contract.test.js:192`, plus Recommended Optional 5 above. Replacement asserts the invariant on the region that states the stop rules.
2. `tests/loop-validate-pipeline-contract.test.js` — the brake regex reads `pr_cycle_count >= N` instead of `review_count >= N`. Case: **Outdated**; the rename is behavior this change deliberately introduces. Ground: Recommended Optional 2 above, which reports the two meanings of `review_count`. Replacement asserts the new name and adds a check that the region states it is never the in-memory `review_count`.

All 293 tests pass (`bun test`).

---
Updated with LLM: Opus 5 | high | Harness: fix-pr-review

richkuo:
@claude review

richkuo:
Addressed review feedback (Opus 4.8 (1M) review via github-actions[bot] · 2026-08-23T23:03:08Z) in 5231ffa, plus a base merge in 9282ecb.

Growth check: diff 555 lines vs 186 at first push (3.0x); cycle 4. The one finding this cycle is a defect in text this PR itself added (`skills/fix-pr-review/SKILL.md:101`, `skills/fix-pr-review/disposition-comment.md:36`), so it takes scope rule 1 and its remedy needs no mechanism this PR lacks.

### Fixed
1. **A reviewer-routed `### Create Follow-up Issue` item the fixer files has no "scope rule it applied" to name in its Deferred-to-follow-up disposition.** — A `Deferred to follow-up` item now names its **basis**, which has exactly two admissible values: the fixer scope rule it applied (rule 2, with the mechanism and the yardstick), or the reviewer's own `### Create Follow-up Issue` routing, where step 4's exclusion filed the item without running a fixer scope rule on it. A reviewer-routed item whose remedy also meets rule 2 may name both (case a). Rule 1 never appears as a basis — it keeps the finding in the PR, so that item is disposed under **Fixed** with the rule-1 note rather than deferred (case b). The existing bullet still routes an already-filed item to its issue number (case c). `pr-review` and all four Action consumers settle a deferral on that basis and require the same treatment to re-raise it (`skills/fix-pr-review/disposition-comment.md:27,36`, `skills/fix-pr-review/SKILL.md:131`, `skills/pr-review/SKILL.md:14`, `templates/claude-workflow/prompts/pr-review-format.md:9`, `templates/codex-workflow/prompts/pr-review-format.md:9`, `templates/claude-review.yml:284`, `templates/codex-review.yml:157`, both fix-pr prompts, `docs/contract-inventory.md:51`).

### Resolved merge conflicts
1. `templates/claude-review.yml` — the base branch rewrote the prior-cycle bullet to pin the fetch command to `${{ github.event.issue.number }}` / `${{ github.repository }}` and to add the untrusted-data instruction, while this branch added the deferral-settle rule to the same paragraph. Resolved by keeping the base's rewritten bullet whole and re-inserting this branch's deferral sentence after "Drop a re-raised finding that carries no such treatment.", so both intents survive. Nine further files auto-merged.

No test edits this cycle — the guards added last cycle were extended to the new wording, never weakened.

All 306 tests pass (`bun test`), including the 13 the base merge brought in.

---
Updated with LLM: Opus 5 | high | Harness: fix-pr-review

richkuo:
@claude sonnet review

richkuo:
Addressed review feedback (Sonnet 5 review via github-actions[bot] · 2026-08-23T23:18:02Z) in 696181a.

Growth check: diff 592 lines vs 186 at first push (3.2x); cycle 5. Re-ran the scope test on the one finding: it is a defect in text this PR itself added (`skills/fix-pr-review/disposition-comment.md:15,36`), so it takes scope rule 1, and its remedy needs no mechanism this PR lacks.

### Fixed
1. **A `Fixed` item that overrides the reviewer's `### Create Follow-up Issue` routing under scope rule 1 has no required field naming that override, unlike the parallel `Deferred to follow-up` basis field this same cycle added.** — The `Fixed` template now carries the field, and a slotting rule requires it in exactly two cases: the reviewer routed the finding to `### Create Follow-up Issue` and step 4's exclusion-exception pulled it back (case a), or the fixer's own rule 2 would have filed the remedy and rule 1 matched first. The note names scope rule 1 and states, from code, what this PR adds, changes, or endangers. Every other `Fixed` item overrides nothing and carries no note — a `Needs Fixing` item, or a `Recommended Optional` item whose remedy needed no mechanism this PR lacks (cases b and c). `pr-review` and all four Action consumers read the note as the mirror of the deferral's basis: the reviewer re-raises its own routing only by showing from current code that rule 1 does not match, and a `Fixed` item that overrode a routing while naming no rule is itself a finding (`skills/fix-pr-review/disposition-comment.md:15,37`, `skills/pr-review/SKILL.md:14`, `templates/claude-workflow/prompts/pr-review-format.md:9`, `templates/codex-workflow/prompts/pr-review-format.md:9`, `templates/claude-review.yml:284`, `templates/codex-review.yml:157`, both fix-pr prompts, `docs/contract-inventory.md:48`).

On case (b) as stated — a rule-1 override on a finding the reviewer routed to `### Recommended Optional` — the rule is written to the substance rather than the section: an optional item the fixer implements is no override at all, because `Recommended Optional` already routes into the PR. The case that does exist there, and that the note now covers, is an optional item whose remedy the fixer's own rule 2 would have filed and rule 1 kept.

No test edits — the guards are additions to existing tests, none weakened.

All 306 tests pass (`bun test`).

---
Updated with LLM: Opus 5 | high | Harness: fix-pr-review

richkuo:
@claude sonnet review

PR #194, opened 2026-08-23
docs(claude-md): let the optimal solution set the tests [C22, Opus 5, high]

richkuo:
## Summary

Replaces the single test sign-off gate in `CLAUDE.md` (symlinked as `AGENTS.md`) with an ordering rule and three self-authorized edit cases.

- **Ordering:** derive the correct behavior from first principles, then bring the tests to it. A failing test is evidence, and the default reading stays that the code is wrong. A test never blocks the right solution.
- **Three cases, no sign-off on any of them:** *outdated* (the change deliberately replaces the asserted behavior), *wrong* (the assertion was never correct), *obsolete* (the behavior is deleted, or another test now covers it).
- **Floor retained:** weakening, deleting, skipping, or narrowing a test whose expectation is still right, to get a green tree, stays forbidden.
- **Disclosure retained and mandatory:** every test edit is named in the commit message and the PR body, with its case and what the replacement asserts.

The mutation-strength requirement from the previous revision is removed.

## Verification

`bun test` — 283 pass, 0 fail across 17 files.

## Plain simple English

The correct solution comes first, and the tests change to match it. You no longer need my approval to fix a test that is outdated, wrong, or no longer needed. You must still not weaken a good test to make a change pass, and you must list every test you changed in the commit and the pull request.

---


richkuo:
@claude review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-08-23T10:33:11Z) in `3bfb195`.

The review was written against head `8cdd7cf`. The policy block was rewritten in `1bd18df` after maintainer directives, so part of the finding's premise no longer holds at head. Line numbers below are at `3bfb195`.

### Corrected scope (partial)

1. **The Obsolete case authorizes deleting a guard test without naming an authority or forcing a strength check, so the weakest-verified path is the one that removes the most protection.** — The redundancy half of the concern is real at head and is now fixed. `CLAUDE.md:48` let an agent drop a guard test on the bare claim that "another test now asserts the same thing", with no check of that claim, and `CLAUDE.md:50` asked only for "what the replacement asserts", which left a removal with no field to fill. `CLAUDE.md:48` now requires the agent to read the surviving test, name it, list each assertion of the removed test that it carries, and keep the removed test when any assertion has no counterpart there. `CLAUDE.md:50` now asks a removal for its ground in place of the replacement.

The two remedies the finding names are ruled out by maintainer directives that postdate `8cdd7cf`:

- **"State who authorizes an Obsolete removal."** `CLAUDE.md:45` now reads "Edit the test yourself in these three cases, and none of them needs my sign-off". The authority question is answered for all three cases uniformly: the agent is authorized. The premise that "line 45 promises the case decides who authorizes it" and that Outdated and Wrong each name a different authority is stale — `1bd18df` removed the per-case sign-off wording, including the explicit sign-off the Wrong case used to require. A sign-off gate on Obsolete would reverse that directive.
- **"Force a mutation proof."** The "at least as strong … proven by mutation" rule the finding cites at old line 49 was removed deliberately in `1bd18df` and must not return in any form.

The gap is therefore closed by tightening what the agent must verify and disclose, which needs no new gate and no mutation run.

**Residual, accepted by directive.** Must-survive case (1) is now caught when the surviving test carries no matching assertion, which is the common shape of a false redundancy claim. A surviving test that does assert the same thing but passes for the wrong reason is still not caught — detecting that needs the mutation proof the maintainer removed. Must-survive case (2), an Obsolete removal of a security or auto-protective test with no human sign-off, is permitted on purpose: no case needs sign-off. The standing guard for both is `CLAUDE.md:49`, which forbids any edit that hides a defect, plus the disclosure at `CLAUDE.md:50`.

Verification: `bun test` — 283 pass, 0 fail. No test file changed.

---

richkuo:
@claude review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-08-23T20:04:41Z) in `393d77d`.

### Fixed

1. **The redundancy check confirms a matching assertion is present in the surviving test, but not that it exercises the removed behavior, so a guard can be dropped for a survivor that passes for the wrong reason.** — Confirmed. "Carries" did read as textual presence at `CLAUDE.md:48` (head `3bfb195`). An assertion now counts as carried only when the counterpart exercises the same behavior: it reaches the real code path rather than a stub or a mock standing in for it, and it can fail on a defect the removed assertion caught rather than restating a constant or a value the test itself set (`CLAUDE.md:48`). This closes all three **Must survive** cases — the mock-backed survivor, the tautological assertion, and the textually similar assertion on a different input class, which fails the "can fail on a defect the removed assertion caught" test. The review's own point stands: this is a static read of the surviving test, which the rule already required, so it adds no sign-off gate and no mutation run and keeps both maintainer directives.

The `3bfb195` disposition's residual is narrowed accordingly. What remains uncaught is the survivor whose assertion does reach the real path and can fail, yet whose failure mode differs from the removed test's in a way no static read reveals. Detecting that needs the mutation proof the maintainer removed by directive.

Verification: `bun test` — 283 pass, 0 fail. CI check `test` green at `3bfb195`.

---

richkuo:
@claude sonnet review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-08-23T20:19:01Z) in `2b0a650`.

### Fixed

1. **The Wrong case authorizes rewriting a test on an unfalsifiable claim that the old expectation was never correct, with no requirement to ground that claim in anything checkable.** — Confirmed. The asymmetry was real at head `393d77d`: Outdated (`CLAUDE.md:46`) and Obsolete's outright-deletion ground (`CLAUDE.md:48`) both trigger on something the diff shows, and Obsolete's redundancy ground was hardened over two cycles to name the surviving test and confirm the counterpart exercises the behavior. Wrong (`CLAUDE.md:47`) carried no anchor at all, and its "say why" was disclosed only after the fact at `CLAUDE.md:50`.

   The claim must now be grounded before the rewrite, with the ground named at disclosure: the code path or contract that contradicts the old expectation, the spec it violates, an issue that reports it, or a maintainer instruction (`CLAUDE.md:47`). When the only support is the agent's reading of what the behavior ought to be, the test stays and the code gets the fix — which routes the ungrounded case back to the section's own default. The disclosure bullet now carries the matching field (`CLAUDE.md:50`), so the three cases are symmetric there.

   This closes all three **Must survive** cases. Case 1 and case 2 both fail the grounding requirement, because new defective or less-secure behavior in the same change is not a source that shows the old expectation was *never* correct. Case 3 gains no advantage, since Wrong is now the more demanding label of the two.

   The review's framing holds: naming a source is a grounding requirement, and it reinstates neither the sign-off gate removed in `1bd18df` nor the mutation proof removed with it.

Verification: `bun test` — 283 pass, 0 fail. CI check `test` green at `393d77d`.

---

richkuo:
@claude review

richkuo:
Addressed review feedback (github-actions[bot] · 2026-08-23T20:45:23Z) in `d3b209b`.

### Fixed

1. **After this PR hardens the Wrong case, the Outdated case becomes the unguarded label: it authorizes rewriting a failing test from the agent's own first-principles reading with no checkable ground, so a regression can be relabeled a deliberate replacement.** — Confirmed, and the diagnosis of the mechanism is right: hardening Wrong at `2b0a650` is what created the asymmetry. "Deliberately replaces" rests on the agent's own judgment, and an accidental regression produces the same diff as an intended replacement, so the diff proves the behavior moved and nothing more.

   The remedy goes one level above the finding's own fix. Patching Outdated alone would repeat the pattern this cycle and the last one both exhibited: harden one case, and the next-weakest becomes the escape hatch. So the burden of proof is lifted into the intro, where it binds every case (`CLAUDE.md:45`) — name a checkable ground before touching the test, independent of your own reading of what the behavior ought to be, and the agent's judgment that a change was deliberate is stated to be the claim under test rather than its own evidence. Each case then names its own ground: Outdated and the deletion branch of Obsolete cite what authorizes the new or deleted behavior (`CLAUDE.md:46`, `CLAUDE.md:48`), Wrong cites what contradicts the old expectation (`CLAUDE.md:47`), and the redundancy branch keeps its surviving-test check (`CLAUDE.md:48`). Outdated also carries the finding's own observation explicitly, so the reason for its ground is visible at the point of use.

   Disclosure now asks for the ground on every case rather than two of three (`CLAUDE.md:50`), which is the finding's second half.

   All three **Must survive** cases close. Case 1 fails for want of an authorizing source, since an accidental regression has none. Case 2 has nowhere to escape to, because every label now owes a ground. Case 3 fails the same way as case 1, and a loosening of a security or auto-protective behavior with no issue or spec behind it leaves the test standing and sends the fix to the code.

   No sign-off gate and no mutation run, so both maintainer directives hold.

Verification: `bun test` — 283 pass, 0 fail. CI check `test` green at `2b0a650`.

---

richkuo:
@claude review

PR #223, opened 2026-08-30
chore(#222): drop MultiEdit from the claude-run.yml allowlists [C6, Fable 5, low]

### LINKED ISSUE #222: [C6] The reusable Claude workflow allows a tool name that the pinned build does not have
**Complexity: 6/100** — Capability 0 (Risk 1 — an allow entry that matches no tool grants nothing); Volume 6 — Sonnet 5, medium · fableplan: no

## Problem

`.github/workflows/claude-run.yml` lists `MultiEdit` in three tool allowlists: the sync-docs default branch at line 132, the implement mode at line 137, and the fix-pr mode at line 140. Every mode runs `anthropics/claude-code-action` at SHA `a874e9e` (lines 203 and 295), which installs Claude Code 2.1.251 per `base-action/action.yml:150`. Issue #198 records that build's answer to a `MultiEdit` rule: `Permission deny rule "MultiEdit" matches no known tool — check for typos.` An allow entry naming an unknown tool grants nothing, so the three entries have no effect. Pull request #220 removed the same name from the review template on that ground, so the repository now states two answers about one build.

## Goal

Every tool name in the `claude-run.yml` allowlists is one the pinned build recognises, so the reusable workflow and the review template agree.

## Approach

1. Remove `MultiEdit` from the three allowlists in `.github/workflows/claude-run.yml` at lines 132, 137, and 140. `Edit` and `Write` stay and cover the file-write surface.
2. Search the repository for any other allow or deny list that names `MultiEdit` and remove it on the same ground.
3. Run `bun test` and parse the workflow with `Bun.YAML.parse`.

## Acceptance criteria

- `grep -rn MultiEdit .github/ templates/` returns nothing.
- `Edit` and `Write` remain in each allowlist that lost `MultiEdit`.
- `bun test` passes and `.github/workflows/claude-run.yml` parses with `Bun.YAML.parse`.

## Plain simple English

The shared bot workflow permits a tool named MultiEdit in three places. The installed bot version has no tool with that name, so the entries do nothing. A recent change removed the same name from the review template. Remove it here too, so both files agree about which tools exist.

---

richkuo:
Closes #222

## Summary

- Removed `MultiEdit` from the three `ALLOWED` lists in `.github/workflows/claude-run.yml` (sync-docs default, implement mode, fix-pr mode). `Edit` and `Write` remain in each.
- The only other `MultiEdit` mention in the repository is the historical record in `docs/contract-inventory.md`, which describes why the name left the review template. It stays, because it is the ground for this change.
- Added a regression test in `tests/pr-review-contract.test.js` that fails when any `claude-run.yml` allowlist names `MultiEdit`. It was run red on the unfixed workflow, then green after the fix.

## Verification

- `grep -rn MultiEdit .github/ templates/` returns nothing.
- `Bun.YAML.parse` reads `.github/workflows/claude-run.yml` without error.
- `bun test`: 362 pass, 0 fail.

No existing test was edited. No implementation plan was posted on the issue.

## Plain simple English

The shared bot workflow listed a tool named MultiEdit in three places. The bot version we pin does not have that tool, so the entries did nothing. This change removes them and adds a test that stops the name from coming back. The workflow and the review template now agree.

---
Created with LLM: Fable 5 | low | Harness: Claude Code

### DIFF
--- a/.github/workflows/claude-run.yml
+++ b/.github/workflows/claude-run.yml
@@ -129,15 +129,15 @@ jobs:
                 ALLOWED='Read,Glob,Grep,Bash(git status*),Bash(git log*),Bash(git tag --sort*),Bash(git fetch*),Bash(git rev-parse*),Bash(gh auth status),Bash(gh release create*),Bash(gh release view*)'
                 ;;
               *)
-                ALLOWED='Read,Edit,Write,MultiEdit,Glob,Grep,Bash(wc *),Bash(ls *),Bash(git status*),Bash(git log*),Bash(git diff*),Bash(git show*),Bash(git fetch*),Bash(git tag --sort*),Bash(git branch*),Bash(git rev-parse*),Bash(git checkout -b docs-sync/*),Bash(git checkout -b docs-release/*),Bash(git add*),Bash(git commit*),Bash(git push origin docs-sync/*),Bash(git push origin docs-release/*),Bash(gh pr create*),Bash(gh pr view*),Bash(gh issue view*)'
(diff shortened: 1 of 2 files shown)


richkuo:
@claude sonnet review

PR #257, opened 2026-09-04
fix(installer): skip already-symlinked skills and workflows [C33, Sonnet 5, high]

richkuo:
## Summary

- `bin/install.mjs` crashed with `ERR_FS_CP_EINVAL` / `ERR_FS_CP_DIR_TO_NON_DIR` whenever a target under `~/.claude/skills` or `~/.claude/workflows` was already a symlink pointing at the matching source in this checkout (the standing dev setup for live-editing skills).
- The installer now detects a target already symlinked to its exact source and leaves it in place instead of copying, reporting it under a new "Left N skills/workflows as-is (already symlinked to this checkout)" line.
- The top summary line ("installed N skills") now counts only skills actually copied, so it no longer claims to install skills it skipped.

## Verification

- Added `bin/install.mjs symlinked source` test in `tests/install.test.js`, seeding a self-referencing symlink and asserting the installer exits 0 and leaves it untouched. Confirmed this test fails against the pre-fix script (same crash) and passes with the fix.
- `bun test`: 312 pass, 0 fail.
- Manually ran `node bin/install.mjs` against the real `~/.claude` (which has ~37 skills and 1 workflow symlinked to this repo): completed cleanly, correctly reported 0 copied / 37 skills and 1 workflow left as-is.

## Plain simple English

The install script failed when a skill link already pointed back to this project folder. This is normal for a developer setup. The script now sees this case and skips the copy, so it no longer crashes and gives an accurate summary.

---
Created with LLM: Sonnet 5 | high | Harness: Claude Code

### DIFF
--- a/bin/install.mjs
+++ b/bin/install.mjs
@@ -1,5 +1,5 @@
 #!/usr/bin/env node
-import { cpSync, existsSync, lstatSync, mkdirSync, readdirSync, renameSync, rmSync } from 'node:fs';
+import { cpSync, existsSync, lstatSync, mkdirSync, readdirSync, readlinkSync, renameSync, rmSync } from 'node:fs';
 import { homedir } from 'node:os';
 import { dirname, join, resolve } from 'node:path';
 import { fileURLToPath } from 'node:url';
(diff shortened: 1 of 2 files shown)


richkuo:
@claude review

PR #312, opened 2026-09-05
docs(#261): shrink the contract inventory to 15765 bytes and cap it at 16000 [C35, Fable 5.1, medium]

### LINKED ISSUE #261: [C35] The contract inventory doc is too large and needs a hard size cap
**Complexity: 35/100** — Capability 1 (Risk 0, Uncertainty 2 — docs only; which rows can be compressed needs discovery); Volume 10 (Scope 1, Coupling 2, Verification 2) — Opus 5, high · fableplan: no

## Problem
`docs/contract-inventory.md` is 38229 bytes, the largest tracked markdown file in the repo (measured with `wc -c` on 2026-09-04; 61 tracked markdown files total 506,293 bytes). It is 103 lines, so most lines are very long paragraphs and wide table rows. It is read by `tests/contract-inventory.test.js` and `tests/first-review-boundary-drift.test.js`, which check its consumer lists against the skill files. Every byte of it is loaded into an agent context window when the file is read, so its size is a direct token cost on every run that uses it.

## Goal
The file is under a hard maximum of 16,000 bytes, states only current behavior, and loses no rule that a test or a consumer depends on.

## Approach
Correctness and safety outrank diff size: no rule that a guard test, workflow, or skill relies on may be dropped.
1. Read the file end to end and list each section with its byte count.
2. For each section decide: keep, compress, move, or delete. Delete text that repeats a rule owned elsewhere and link to the owner instead. Move history and rationale out of the file. Compress long prose into tables or single-line rules.
3. Check every remaining claim against the current code and skills; correct or remove anything the codebase contradicts.
4. Enforce the cap: add a byte-size assertion to `tests/contract-inventory.test.js` that fails when the file exceeds 16,000 bytes.
5. Run `bun test` and confirm every existing guard still passes with no test weakened.

## Acceptance criteria
- `wc -c docs/contract-inventory.md` reports under 16,000 bytes.
- A size guard test fails when the file grows past 16,000 bytes.
- No consumer named above loses a rule it depends on; `bun test` passes with no test edited, or every test edit is disclosed with its ground.
- No claim in the file contradicts the current code or skills.

## Plain simple English
The document that lists which skills share which pipeline rules has grown to about 38 kilobytes. It is the largest text file in the repo. Agents load it into memory on each run, so the size costs tokens. It must be cut to under 16 kilobytes and a test must stop it from growing again.

---
Created with LLM: Fable 5.1 | high | Harness: Claude Code

richkuo:
Closes #261

## Summary

- `docs/contract-inventory.md` goes from 38229 to 15765 bytes (`wc -c`). Each rule row now states only the markers its guard checks and points at the owner file for the full wording; the skill table is a grouped list; the exception, out-of-scope, and related-coverage sections are condensed; the eleven stacked attribution lines collapse to one.
- Every repository path and guard test the old file named is kept. The band-derived trigger row now points at `skills/fix-pr-review/rereview-routing.md` for the re-review ladder, which the old row attributed to `fix-pr-review` alone.
- `tests/contract-inventory.test.js` gains a size guard that fails when the file reaches 16,000 bytes.

## Verification

- `bun test`: 315 pass, 0 fail, 19 files.
- The new size guard was run against the pre-change file (checked out from `HEAD`) and failed; it passes on the new file.
- `tests/first-review-boundary-drift.test.js` still pins the boundary restatements in this file to the `validate-issue` step 6 table.

## Test edits

- `tests/contract-inventory.test.js`: one new test added, no existing assertion changed. Ground: issue #261 acceptance criterion "A size guard test fails when the file grows past 16,000 bytes". It asserts `Buffer.byteLength` of the file is below 16,000.

## Plan

No plan was posted on the issue; deviations: none.

## Plain simple English

The document that lists the shared pipeline rules was 38 kilobytes. It is now under 16 kilobytes. Each row keeps the part a test checks and points to the file that owns the full rule. A new test fails if the file grows past 16 kilobytes again.

---
Created with LLM: Fable 5.1 | medium | Harness: work-on-issue

https://claude.ai/code/session_01V8TZ4Kw38DZ3W913Pvtops


richkuo:
@claude review

PR #317, opened 2026-09-07
fix(fix-pr-review): make review repairs complete and resumable [C66, GPT-6, high]

richkuo:
## Summary

Rework feedback tracking, finding status, branch safety, and publication retries; align both workflow prompts.

## Verification

319 tests passed; skill validation, reference checks, GraphQL query, and four scenario simulations passed. No tests changed.

## Plain simple English

Unfinished findings stay visible, and repeated runs check what was already posted.

---
Updated with LLM: GPT-6 | high | Harness: Claude Code


richkuo:
@claud fable review effort:high

richkuo:

richkuo:
Addressed review feedback (@claude fable review, 2026-09-07T20:20:07Z) in af621cab48397a5b461312100767a5ceb031bbd4.

### Fixed
1. **The Claude fix-pr route cannot run the REST commands the skill files it now points at prescribe, so inline replies never post and the pass ends unpublished.** The Claude route now stages the skills directory (`.github/workflows/claude-run.yml:75`), names the staged path in the prompt (`:233`), passes `--add-dir` for it (`:240`), and admits `gh issue list` (`:177`). The Claude prompt fetches with `gh pr view --json reviews,comments` and GraphQL (`templates/claude-workflow/prompts/fix-pr.md:5`) and replies to inline threads with the `addPullRequestReviewThreadReply` mutation (`:15`), so the route needs no REST call. The allowlist was not widened to `gh api repos/*`, because that prefix also admits write and delete endpoints. Regression test: `tests/pr-review-contract.test.js:200`.
2. **A "blocked pass" is undefined, and its no-trigger outcome leaves the loop caller with no next step.** A pass is now blocked when any finding ends Blocked, a stop before step 8 leaves no commit, or the push or publication did not complete (`skills/fix-pr-review/SKILL.md:95`). A blocked pass pushes and publishes its completed work and posts no trigger (`:101`). The skill reports one of three outcomes to the loop: complete, pending trigger, or blocked. `skills/fix-pr-review-loop/SKILL.md:57` routes blocked to a new **Blocked** row (`:71`) and waits on a pending trigger without counting it. `skills/work-on-issue-loop/SKILL.md:39` matches. Both prompts carry the definition (`templates/*/prompts/fix-pr.md` Phase 8). Regression test: `tests/pr-review-contract.test.js:218`. Scope rule 1 applies: this PR added the Blocked state. The remedy differs from the proposed one; see the judgment call below.
3. **The Opus rung rule differs between the skill and the Claude prompt.** The Claude prompt now posts the Opus rung until an Opus review has run. While an Opus request has no review, the prompt posts no new trigger and names that request in the disposition. The fallback row now keys on an Opus review that ran (`templates/claude-workflow/prompts/fix-pr.md:17`). This matches `skills/fix-pr-review/rereview-routing.md:37` and `:53`.
4. **Attribution footers are now embedded in the two prompts and five skill files, which no other prompt or skill carries.** Removed from all seven files. Prompt sizes are now 9859 bytes (Claude) and 9942 bytes (Codex), under the 10000-byte cap.
5. **The contract inventory still names `red-flags-and-mistakes.md` as a consumer of the broken-test rule this PR removed from it.** Removed that file from the row (`docs/contract-inventory.md:31`). `skills/fix-pr-review/SKILL.md` step 6 owns the rule.

### Resolved judgment calls
1. **A "blocked pass" is undefined, and its no-trigger outcome leaves the loop caller with no next step.** Implemented: any Blocked finding makes the pass blocked, and a blocked pass posts no trigger and stops the loop. Alternative rejected: publishing and triggering when some findings are Blocked. A re-review raises the Blocked item again, the loop runs the fixer, the item stays Blocked, and a trigger is posted again. No stop rule catches that sequence, because `Needs Updates` never stops the loop on cycle count alone (`skills/fix-pr-review-loop/SKILL.md:53`). Cost: fixes pushed next to a Blocked item wait for review until the maintainer supplies the missing decision.

### Verification
`bun test`: 321 pass, 0 fail. No existing test was edited. Two tests were added.

## Plain simple English
The Claude workflow can now read the skill files and uses only commands that it has permission to run. A pass that needs a human decision now stops and tells the loop to stop. The skill and the prompt now use the same Opus rule. The wrong signature lines are gone.

---
Created with LLM: Opus 5.5 | high | Harness: Claude Code

richkuo:
@claude opus review effort:high

richkuo:
@claude opus review effort:high

richkuo:
Closed in favor of #334, which carries only the low-risk parts of this PR (thread reply pagination, older-review reconciliation, push and retry safety). The Blocked state and the full rewrite caused new gaps in each review cycle, so they are left out.

---
Updated with LLM: Opus 5.5 | high | Harness: Claude Code

PR #319, opened 2026-09-07
docs(validate-issue): rework evidence and update checks [C68, GPT-6, high]

richkuo:
## Summary

Issue validation could use the wrong repository, treat incomplete evidence as a completed result, or overwrite newer issue text. The revised skill pins evidence to a verified commit, stops incomplete validation, and rechecks issue content before saving corrections.

The review removes repeated architecture and consistency checks, duplicated update rules, shared style copies, the fixed history window, and reply menus. It adds observable acceptance checks, explicit planning for missing scores and safety findings, and a static-inspection verification grade. The score formula, routing tables, numbered entrypoints, and completed-verdict fields remain compatible with callers.

## Verification

- Full Bun suite: 319 tests passed. Final focused contract checks also passed.
- Both changed skill entrypoints pass quick_validate.
- A temporary Git fixture confirmed that recorded-commit reads ignore divergent, staged, dirty, deleted, and untracked files, remain stable after a branch reference moves, and preserve repository status.
- Relative Markdown links, numbered entrypoints, attribution, and git diff --check passed.
- GitHub issue writes were not exercised. The freshness check reduces stale overwrites but does not provide an atomic concurrency guarantee.

## Test edits

- **complexity-score.test.js: Outdated.** Ground: the user's request to remove repetition. Both rescore tests now read the canonical editing reference and retain checks for synchronized scores, upward-only routing, protected harness stamps, and the effective planning signal.
- **issue-plain-simple-english-contract.test.js: Outdated.** Ground: the user's request and CLAUDE.md Response Style requirement to reference shared rules. The cap check accepts an explicit owner reference and verifies the owner's cap; body-order and missing-section checks remain.
- **loop-validate-pipeline-contract.test.js: Wrong.** Ground: CLAUDE.md requires Updated for edited artifacts, while issue validation requires Validated. The test separates actual terminal artifact attribution from procedure text, retains generated issue-footer checks, and checks the editing reference's own Updated footer.

## Plain simple English

The skill now checks the correct code snapshot, separates missing evidence from a valid result, and checks for newer issue text before saving changes. Repeated instructions were removed. Existing scoring and workflow formats remain available to the related skills.

---
Updated with LLM: GPT-6 | high | Harness: Codex

### DIFF
--- a/skills/fable-validate/SKILL.md
+++ b/skills/fable-validate/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: fable-validate
-description: Use when the user wants a GitHub issue validated by a Fable 5.1 subagent. Spins up a read-only subagent running on Fable 5.1 that executes the validate-issue procedure (claim tracing, architecture/consistency checks, complexity score), then relays the verdict back to the main agent, which presents it and takes any follow-on action (update issue, work on issue). Trigger on "/fable-validate", "fable validate <issue>", or "validate this with fable".
+description: Use when the user wants a GitHub issue validated by a Fable 5.1 subagent. Spins up a read-only subagent running on Fable 5.1 that executes the validate-issue procedure (claim tracing, architecture/consistency checks, complexity score), then relays the verdict back to the main agent, which presents it and takes any follow-on action (update issue, work on issue). Trigger on "/fable-validate", "fable validate an issue", or "validate this with fable".
 ---
 
 # fable-validate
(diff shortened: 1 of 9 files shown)


richkuo:

richkuo:
Request: make validate-issue always land a complexity score on the issue, even when the original issue has no `[C<score>]` prefix.

This pull request keeps the current rule in both places: "a title with no prefix gets none from a rescore" (SKILL.md step 8 and issue-editing.md), and the missing-score fallback that routes unscored issues at the highest band forever (complexity-scoring.md). The validator computes a full score either way, but with no prefix it never lands on the issue, so an unscored issue keeps top-band routing at validate, build, and review no matter how simple the work is.

Concrete case: richkuo/go-trader#1544 was filed by an external tool with no prefix. Validation traced 64/100 (band 3). The skill blocked stamping it, so the issue would have kept highest-band routing permanently. I stamped it only because the user explicitly overrode the rule.

Proposed change: when validation ends in Update yes, add the recomputed prefix and rationale line even when no prefix exists. Keep "never lower routing" scoped to issues that already carry a score. The highest-band fallback then applies only before the first validation, which is the only window where no traced score exists.

---
Created with LLM: Kimi | high | Harness: Kimi Code

richkuo:
Follow-up to the request above: it is also acceptable for validate-issue to lower the band when the traced score comes out below the title score.

A primary objective of validate-issue is to score the issue accurately against the codebase, covering the cases where the original issue is inaccurate, outdated, or already partially fixed. The current "never lower routing from a validator rescore" rule blocks exactly that correction: an over-scored title keeps its high band permanently, so build and review spend top-model effort that the traced complexity does not justify.

Proposed change: let a validation rescore restamp downward as well as upward, in both the title prefix and the rationale line, with the evidence for the lower grades recorded in the verdict. The protection against accidental downgrades should come from the grading evidence and the Differs lines, not from a floor that freezes the first score forever.

---
Created with LLM: Kimi | high | Harness: Kimi Code

richkuo:
Superseded by #336, which rebuilds the useful parts of this pull request on current main (evidence pinned to one commit, a blocked outcome that callers treat as a stop, and a freshness check before an issue save) and applies both owner comments above (stamp a score on unscored issues, and allow a lower score on evidence). This branch conflicts with main and its review findings stay open, so it closes without merge.

---
Created with LLM: Opus 5.5 | high | Harness: Claude Code

